PR #7, opened 2021-01-16
remove A1/A2 <-> M1/M2 channel aliasing

jrevels:
It turns out that these are not the same, though they are similarly used as references. 

From @ararslan: "M1/2 are the mastoids and A1/2 are earlobes."


palday:
Strictly speaking following the 10-20/10-10 systems, this is true. But I've seen A1/A2 used for M1/M2 in many real labs (i.e. true mastoid electrodes aliased to A1/A2). And I'm guessing that's where the confusion comes from. That also means that if we're going to be aliasing at any point, it should be A corrected/aliased to M. (I've never seen M used for true A.)  It's rare but possible to have both earlobes and mastoids, so not aliasing (by default) is probably ideal. But then it might make sense to have an easy way to correct "known" mislabeling. 


(Also note to future self: the 10/20 names for the EOG - `SO[1,2]`, `IO[1,2]`, `LO[1,2] ` electrodes aren't listed here, nor are `VEOG` and `HEOG`, which are common names for bipolar ocular channels.)

ararslan (approved):

ararslan:
Tests need to be adjusted for this change, but the implementation of the change itself LGTM

ararslan (approved):

PR #15, opened 2021-04-20
Also allow Onda v0.13

kleinschmidt:
I think this is pretty straightforwardly compatible with Onda v0.13 and all tests pass but please tell me if I'm missing something.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OndaEDF"
 uuid = "e3ed2cd1-99bf-415e-bb8f-38f4b42a544e"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.8.0"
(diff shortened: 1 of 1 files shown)


ararslan (approved):
Should be fine, as https://github.com/beacon-biosignals/Onda.jl/compare/v0.12.0...v0.13.0 seems like a collection of generally non-breaking things

kleinschmidt:
Yeah that was my thought too.  Is it just because the changes to the arrow timespan serialization?

jrevels:
> Is it just because the changes to the arrow timespan serialization?

IIRC it's actually because the `materialize` kwarg was deprecated from a couple of API functions, ref https://github.com/beacon-biosignals/Onda.jl/pull/71/files

PR #18, opened 2021-04-20
empty annotations when not importing annotations

palday:
An hour has elapsed: time for a patch release!

### DIFF
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -231,7 +231,7 @@ in `\$path/samples/`, and write the Onda signals and annotations tables to
 "edf", and if a prefix is provided for signals but not annotations both will use
 the signals prefix.  The prefixes cannot reference (sub)directories.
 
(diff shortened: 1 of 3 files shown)


src/import_edf.jl
```diff
@@ -231,7 +231,7 @@ in `\$path/samples/`, and write the Onda signals and annotations tables to
 "edf", and if a prefix is provided for signals but not annotations both will use
 the signals prefix.  The prefixes cannot reference (sub)directories.
 
-Returns `uuid => (signals, annotations)`.
+Returns `uuid => (signals, annotations)`.
```
palday, line 234:
```suggestion
Returns `uuid => (signals, annotations)`.
```

kleinschmidt (changes requested):
good catch.  can you add a test case for this?

kleinschmidt (approved):

PR #19, opened 2021-04-20
Try, try again (to build release docs)

kleinschmidt:
Copied in a workflow that is known to work for building released docs (https://github.com/JuliaStats/StatsModels.jl/blob/master/.github/workflows/docs.yml)


christopher-dG (approved):
This *shouldn't* do anything, because the triggers are exactly the same, but it's worth a try!

PR #22, opened 2021-04-21
Document and lightly refactor signal extraction during import

kleinschmidt:
This documents and refactors some internal functions to make it slightly easier
to create custom extractor functions.  The main change is that the construction
of the `Onda.SamplesInfo` from a collection of `EDF.Signal`s is extracted into
its own function, since that's likely to be needed by custom extractors that
can't rely on the labels.  Additionally, this adds docstrings for
`extract_channels_by_label` and the new `edf_signals_to_samplesinfo` functions.

It does not export them or include them in the docs, although that may be the
right way to go if they're now meant to be used to create custom extractors.

### DIFF
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -204,7 +204,7 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
 a canonical name to a list of alternatives that it should be substituted for
 (e.g., `"canonical_name" => ["alt1", "alt2", ...]`).
 
-See OndaEDF.STANDARD_LABELS for the labels (`signal_names => channel_names`
+See `OndaEDF.STANDARD_LABELS` for the labels (`signal_names => channel_names`
 `Pair`s) that are used to extract EDF signals by default.
 
 """
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -163,7 +163,7 @@ function extract_channels(edf_signals, channel_matchers)
 end
 
 """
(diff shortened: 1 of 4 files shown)


ararslan:
> It does not export them or include them in the docs, although that may be the right way to go if they're now meant to be used to create custom extractors.

I'd document and export.

kleinschmidt:
To be clear, they are documented in the sense that they have docstrings, just didn't include them in the HTML docs

ararslan:
Right, I mean include in the docs

ararslan (approved):

src/import_edf.jl
```diff
@@ -162,6 +162,52 @@ function extract_channels(edf_signals, channel_matchers)
     return extracted_channel_names, extracted_channels
 end
 
+"""
+    edf_signals_to_samplesinfo(edf, edf_signals, kind, channel_names, samples_per_record)
+
+Generate a single `Onda.SamplesInfo` for the given a collection of `EDF.Signal`s
```
jrevels, line 168:
```suggestion
Generate a single `Onda.SamplesInfo` for the given collection of `EDF.Signal`s
```

src/import_edf.jl
```diff
@@ -162,6 +162,52 @@ function extract_channels(edf_signals, channel_matchers)
...
+corresponding to multiple channels from a single signal kind.  Sample units are
```
jrevels, line 169:
```suggestion
corresponding to the channels of a single Onda signal.  Sample units are
```

jrevels (approved):

PR #23, opened 2021-05-18
update `STANDARD_{LABELS,UNITS}` and label parsing to accommodate seen-in-the-wild recordings

kolia:
These changes allow for certain signal headers that are not  up to the EDF+ spec to be nonetheless imported, because they are very common and un-ambiguous to the human interpreter.

In particular:
- the spec states labels should have the form `<type><some spaces><label>`; many signal labels do not start with the signal type: the spec-compliant `EEG C3-O1` is often just `C3-O1`, which was ok as well previously, but even when the type is absent, spaces are often present in the label, which throws off what the standard suggests doing, which is split on space to get the type; this attempts to fix that.
- sometimes the information that should be in the `label` field is in the `transducer type` instead; this first tries using the label, then tries using the transducer type as if it were a label.
- this allows passing in a `preprocess_labels` that takes a label string in to modify.
- some non-ACSII characters appear in the signal headers; some of these are malformed UTF-8 which were triggering errors, which this PR deals with putting a try catch around the calls to `lowercase` that were failing; using `preprocess_labels` with simple character replacements can help, as is done in `test_preprocessor`.

Whether each of these non-EDF+-compliant updates should be in  `OndaEDF.jl` is up to debate. This PR takes the pragmatic approach of:
- adding tests based on a long list pairs matching EDF headers as parsed by `EDF.jl` with their expected Onda `SamplesInfo`, that includes many non-EDF+-compliant pairs, designed to capture a variety of EDFs seen in the wild
- making it easier to customize extractors by
   - making the the units that default to `STANDARD_UNITS` passable as kwargs
   - using `preprocess_labels` to massage labels before extraction
   - making custom extractors replace the default extractors, instead of being appended to them as previously.
- trying to accommodate non-EDF+-compliant signal headers if they are un-ambiguous, using the 3 custom extractor hooks would be awkward, and the change can be done without painful cringing


test/import.jl
```diff
@@ -1,7 +1,46 @@
...
+    end
+    # "C2M1" => "C2-M1" etc
+    m = match(r"\s*(?<channel>[fco][1234])\s*(?<ref>[am][12])\s*"i, l)
+    isnothing(m) && return l
+    return "$(m[:channel])-$(m[:ref])"
+end
+
+custom_extractors = [edf -> extract_channels_by_label(edf, signal_names, channel_names; preprocess_labels=test_preprocessor)
+                     for (signal_names, channel_names) in OndaEDF.STANDARD_LABELS]
+
```
kolia, line 26: For tests, use `preprocess_labels` appropriate for `test_edf_to_samples_info.out`.

test/runtests.jl
```diff
@@ -14,6 +13,7 @@ function test_edf_signal(rng, label, transducer, physical_units,
     return EDF.Signal(header, samples)
 end
 
+
```
kolia, line 16:
```suggestion
```

test/signal_labels.jl
```diff
@@ -6,7 +6,7 @@
     @test OndaEDF.match_edf_label("EEG C3-(M1 +A2)/2 - rEf3", signal_names, "c4", canonical_names) == nothing
     @test OndaEDF.match_edf_label(" TEsT   -Fpz  -REF-cpz", signal_names, "fpz", canonical_names) == "-fpz-ref-cpz"
     @test OndaEDF.match_edf_label(" TEsT   -Fpz  -REF-cpz", signal_names, "fp", canonical_names) == nothing
-    @test OndaEDF.match_edf_label("  -Fpz  -REF-cpz", signal_names, "fpz", canonical_names) == nothing
+    @test OndaEDF.match_edf_label("  -Fpz  -REF-cpz", signal_names, "fpz", canonical_names) == "-fpz-ref-cpz"
```
kolia, line 9: New heuristic says that this test case is acceptable. EDF standard doesn't rule one way or the other afaiu.

src/import_edf.jl
```diff
@@ -48,17 +40,29 @@ function _normalize_references(original_label, canonical_names)
...
-        label = type
-    else
-        any(==(type), signal_names) || return nothing
-        label = spec
+    label = _safe_lowercase(label)
+    for signal_name in signal_names
+        m = match(Regex("[\\s\\[,\\]]*$(signal_name)[\\s,\\]]*\\s+(?<spec>.+)", "i"), label)
+        if !isnothing(m)
+            label = m[:spec]
+        end
```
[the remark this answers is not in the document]
kleinschmidt: won't this fail on cases described in these deleted lines? https://github.com/beacon-biosignals/OndaEDF.jl/pull/23/files#diff-0ec7e29ad703d1c057db3e96603f12046612e73f6f652cef6d64230fd2adfe52L11-L12

i.e. where there's no separate spec...

oh I see, you're just replacing label with the spec part of the regex, and if the regex doesn't matches you just use the whole label as before.  might be worth moving that deleted comment down here to clarify what's going to get picked up.

src/import_edf.jl
```diff
@@ -163,7 +167,7 @@ function extract_channels(edf_signals, channel_matchers)
 end
 
 """
-    edf_signals_to_samplesinfo(edf, edf_signals, kind, channel_names, samples_per_record)
+    edf_signals_to_samplesinfo(edf, edf_signals, kind, channel_names, samples_per_record; unit_alternatives=STANDARD_UNITS)
```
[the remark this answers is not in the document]
kleinschmidt: I wonder if it makes sense to always use the standard units and allow for additional custom overrides/additions?  I guess then you'd have to merge the unit sets which might be annoying.

kolia:
Looking at some of the ntuples in `test/test_edf_to_samples_info.out` should be informative, as all of the (EDF.Signals, Onda.SamplesInfo) pairs reflect what this PR does.

I have paid much more attention to the outcomes for EEG than for other signal types.

src/import_edf.jl
```diff
@@ -172,10 +176,12 @@ converted to Onda units and checked for consistency, and a promoted encoding
...
-function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Signal}, kind, channel_names)
-    onda_units = map(s -> edf_to_onda_unit(s.header.physical_dimension), edf_signals)
-    onda_sample_unit = first(onda_units)
-    all(==(onda_sample_unit), onda_units) || error("multiple possible units found for same signal: $onda_units")
+function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Signal}, kind, channel_names; unit_alternatives=STANDARD_UNITS)
+    onda_units = map(s -> edf_to_onda_unit(s.header.physical_dimension, unit_alternatives), edf_signals)
+    onda_sample_units = Set(u for u in onda_units if u != "unknown")
+    onda_sample_units = isempty(onda_sample_units) ? Set(["unknown"]) : onda_sample_units
+    length(onda_sample_units) == 1 || error("multiple possible units found for same signal: $onda_sample_units")
+    onda_sample_unit = first(onda_sample_units)
```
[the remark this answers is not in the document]
kleinschmidt: Add something to the docstring about this behavior?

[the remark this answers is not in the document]
kleinschmidt: I'm not sure this behavior is desirable in practice...it seems like a major potential footgun to silently and automatically assume that, say, an EMG signal is in microvolts just because the other EMG signals are since EMG signals are also often in volts...  I don't know enough about these setups to know whether it's even possible that you'd get multi-unit setups for the same signal type AND have unknown units on top of that, so may be worth checking with some domain experts.

The alternative would be to error (the existing behavior) or to discard the "unknowns".

src/import_edf.jl
```diff
@@ -189,6 +192,7 @@ function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Sig
     return info
 end
 
+
```
kleinschmidt, line 195:
```suggestion
```

src/import_edf.jl
```diff
@@ -204,22 +208,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
 a canonical name to a list of alternatives that it should be substituted for
 (e.g., `"canonical_name" => ["alt1", "alt2", ...]`).
 
+`unit_alternatives` lists standardized unit names and alternatives that map to them.
+See `OndaEDF.STANDAR_UNITS` for defaults.
```
kleinschmidt, line 212:
```suggestion
See `OndaEDF.STANDARD_UNITS` for defaults.
```

src/import_edf.jl
```diff
@@ -189,6 +192,7 @@ function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Sig
     return info
 end
 
+
 """
     extract_channels_by_label(edf::EDF.File, signal_names, channel_names)
```
kleinschmidt, line 197:
```suggestion
    extract_channels_by_label(edf::EDF.File, signal_names, channel_names; unit_alternatives=STANDARD_UNITS, preprocess_labels=identity)
```

src/import_edf.jl
```diff
@@ -204,22 +208,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
...
+function extract_channels_by_label(edf::EDF.File, signal_names, channel_names; unit_alternatives=STANDARD_UNITS, preprocess_labels=identity)
     matcher = x -> begin
         # yo I heard you like closures
-        return s -> match_edf_label(s.header.label, signal_names,
-                                    x isa Pair ? first(x) : x,
-                                    channel_names)
+        return s -> begin
+            m = match_edf_label(preprocess_labels(s.header.label),
+                                signal_names,
+                                x isa Pair ? first(x) : x,
```
kleinschmidt, line 227: Now that this is in two places, could you lift this logic out with a comment explaining what `x` is, why it might be a pair, and why, if it is, `match_edf_label` needs the `first` of it?

src/import_edf.jl
```diff
@@ -204,22 +208,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
...
+            # channel info is sometimes misplaced in transducer_type field; only accept these if corresponding signal_name occurs in channel label
+            m = match_edf_label(preprocess_labels(s.header.transducer_type),
+                                signal_names,
+                                x isa Pair ? first(x) : x,
+                                channel_names)
+            isnothing(m) && return m
+            lowercase_label = _safe_lowercase(s.header.label)
+            any(occursin(name, lowercase_label) for name in signal_names) && return m
+            return nothing
+        end
```
kleinschmidt, line 239: fix isnothing check and lift out/comment on the x isa Pair thing:

```suggestion
        # x is either a channel name (string), or a channel_name => alternatives Pair.
        this_channel_name = x isa Pair ? first(x) : x
        return s -> begin
            m = match_edf_label(preprocess_labels(s.header.label),
                                signal_names,
                                this_channel_name,
                                channel_names)
            !isnothing(m) && return m
            # channel info is sometimes misplaced in transducer_type field; only accept these if corresponding signal_name occurs in channel label
            m = match_edf_label(preprocess_labels(s.header.transducer_type),
                                signal_names,
                                this_channel_name,
                                channel_names)
            !isnothing(m) && return m
            lowercase_label = _safe_lowercase(s.header.label)
            any(occursin(name, lowercase_label) for name in signal_names) && return m
            return nothing
        end
```

src/import_edf.jl
```diff
@@ -204,22 +208,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
...
+                                channel_names)
+            !isnothing(m) && return m
+            # channel info is sometimes misplaced in transducer_type field; only accept these if corresponding signal_name occurs in channel label
+            m = match_edf_label(preprocess_labels(s.header.transducer_type),
+                                signal_names,
+                                x isa Pair ? first(x) : x,
+                                channel_names)
+            isnothing(m) && return m
+            lowercase_label = _safe_lowercase(s.header.label)
+            any(occursin(name, lowercase_label) for name in signal_names) && return m
```
kleinschmidt, line 237: can you explain the situations where this fallback is appropriate?  and given that the logic of this closure has become more complex, maybe it should be rolled into `match_edf_label` or another helper function?

src/import_edf.jl
```diff
@@ -365,10 +396,10 @@ following transformations:
 
 See the OndaEDF README for additional details regarding EDF formatting expectations.
 """
-function edf_to_onda_samples(edf::EDF.File; custom_extractors=())
+function edf_to_onda_samples(edf::EDF.File; custom_extractors=STANDARD_EXTRACTORS)
     EDF.read!(edf)
     edf_samples = Samples[]
-    for extractor in Iterators.flatten((STANDARD_EXTRACTORS, custom_extractors))
+    for extractor in custom_extractors
```
kleinschmidt, line 402: can you explain the motivation for making custom_extractors replace instead of extend the standards?  I thought in a previous version of this PR you'd prepended them instead of postpending, which seems like a better choice to me but something clearly made you change your mind.

kleinschmidt (commented):
I think technically this looks fine (I noticed one place with `isnothing && return` that seems like it shoudl be `!isnothing && return`).  As far as the design goes, that's more of a judgement call as you point out.  I think it would be good to add some documentation about the conditions under which we'll consider weird edge cases people encounter in the wild to this package vs. say "use a custom extractor".  That's always going to be a judgement call but I think we need at least some minimal guidelines about how and why we'd consider extending the standards, especially once we've gone "out of bounds" of the spec.

src/import_edf.jl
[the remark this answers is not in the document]
kolia: I figure anyone passing in custom units will have looked at the source code (the only place where the format of units dict is currently documented..) so they can cut-and-paste and modify the current standard ones.

src/import_edf.jl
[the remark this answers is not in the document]
kolia: An alternative design could be to define Errors containing fields that capture what the error is about (so for example, the vector of parsed channels and corresponding units), and letting users pass in an `on_error` callable that handles these, either by re-throwing or by returning something. I kinda like that.

src/import_edf.jl
kolia: Using the `transducer_type` as a label is a little dangerous, because all kinds of garbage can show up in the `transducer_type` field which could spuriously match some signal channels; this only accepts a successful match using the transducer type if the label mentions the corresponding signal type.

This whole `transducer_type -> label` fallback is the messiest part of this PR, I don't like it. The main reason I put it there is because it would be annoying to provide a custom extractor to do this, unless I'm overlooking some appropriate design. Lemme count how many test EDF signals would be affected without this.

kolia:
Much has changed since trying to accomodate `E{MOK}G` as well as `EEG`.

Closing this and opening a new PR against the same branch, since the comments here are outdated.

tia-volkamer:
Closed as outdated duplicate of #24

PR #24, opened 2021-06-01
update default extractors + `STANDARD_{LABELS,UNITS}` to accommodate seen-in-the-wild recordings

kolia:
This PR is inspired by ingesting a dataset collected from 80+ sites, with small but significant variations within sites, and large variations between. See `test/test_edf_to_samples_info.in` for a taste. (This file contains all the unique sets of EDF signal headers, each set of headers typically maps to many recordings.)

These changes allow for more signal labels to be importable. Some spec-compliant cases that previously threw errors are now dealt with. More forms of non-compliance with the EDF+ spec are imported than previously, which are very common and (most often) un-ambiguous to the human interpreter.

This PR takes the pragmatic approach of:
- adding tests based on a long list pairs matching EDF headers as parsed by `EDF.jl` with their expected Onda `SamplesInfo`, that includes many non-EDF+-compliant pairs, designed to capture a variety of EDFs seen in the wild
- having these tests write pretty-printed julia capturing extraction results to `test/test_edf_to_samples_info.out`, which can be diffed with existing/expected `test/test_edf_to_samples_info.in` to see the effect of any code changes
- write subsets of `test/test_edf_to_samples_info.out` to `test/no_e{eko}g.out`, `test/no_leg.out` and `test/no_chin.out`, which are subsets of results for which the corresponding signal type is not present, to make it easier to check that extraction is not missing signals.
- making it easier to customize extractors by
   - making the units that default to `STANDARD_UNITS` customizable as a kwarg
   - using `preprocess_labels(label, transducer_type)` to massage labels before extraction; its output will be used instead of the original `label` field
   - making custom extractors replace the default extractors, instead of being appended to them as previously.

In particular:
- the spec states labels should have the form `<type><some spaces><label>`; many signal labels do not start with the signal type: the spec-compliant `EEG C3-O1` is often just `C3-O1`, which was ok as well previous to this PR, but even when the type is absent, spaces are often present in the label, which throws off what the standard suggests doing, which is split on space to get the type; this PR attempts to fix that.
- previously, the same `EDF.Signal` could be extracted into multiple signal channels without warning; this PR excludes such ambiguous signals, and turns them into `AmbiguousChannelError`s that are not thrown but get returned alongside extracted `Onda.Samples`.
- signals of the same type but with different physical units or sample rates are extracted as multiple separate signals that have the same type; previously an error was thrown.
- some non-ACSII characters that appear in the signal headers can be malformed UTF-8 which were triggering errors; this PR deals with them by putting a try catch around the calls to `lowercase` that were failing; using `preprocess_labels` with simple character replacements can help, as is done in `test_preprocessor`.
- this adds a `emg_ambiguous` signal type, so that ambiguous EMG channels can be disambiguated during post-processing that takes all extracted channels into account (not part of this code); typically a channel with a label similar to `EMG 1` can be interpreted as a chin channel if leg EMG channels are also present, otherwise it is ambiguous. `emg_ambiguous` appears in 171 out of our 992 wild test cases.

Given that the see-in-the-wild EDF headers in `test/test_edf_to_samples_info.in` are very often far from the EDF+ standard, a good amount of oddball massaging is done to them in `test_preprocessor` defined in `test/import.jl`. **Whether each of these non-EDF+-compliant fixes should be part of `OndaEDF.jl`'s default extractors is up to debate.** I tried to only accommodate non-EDF+-compliant signal headers in the default extractors if:
   - they are un-ambiguous
   - using the 3 custom extractor hooks would be awkward
   - and the change can be done without cringing


kolia:
For reviewers, I recommend first taking a look at `test/test_edf_to_samples_info.out` to get a sense for the variety of EDF Signal headers seen and what Onda channels this PR maps them to.

src/import_edf.jl
```diff
@@ -189,12 +189,30 @@ function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Sig
...
 end
 
+struct SamplesInfoError <: Exception
+    msg::String
+    cause::Exception
+end 
+
+function groupby(f, list)
+    d = Dict()
+    foreach(list) do v
```
kleinschmidt, line 199:
```suggestion
    for v in list
```

src/import_edf.jl
```diff
@@ -204,22 +222,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
 a canonical name to a list of alternatives that it should be substituted for
 (e.g., `"canonical_name" => ["alt1", "alt2", ...]`).
 
+`unit_alternatives` lists standardized unit names and alternatives that map to them.
+See `OndaEDF.STANDAR_UNITS` for defaults.
```
kleinschmidt, line 226:
```suggestion
See `OndaEDF.STANDARD_UNITS` for defaults.
```

src/import_edf.jl
```diff
@@ -189,12 +189,30 @@ function edf_signals_to_samplesinfo(edf::EDF.File, edf_signals::Vector{<:EDF.Sig
...
 
+struct SamplesInfoError <: Exception
+    msg::String
+    cause::Exception
+end 
+
+function groupby(f, list)
+    d = Dict()
+    foreach(list) do v
+        push!(get!(d, f(v), []), v)
```
kleinschmidt, line 200: I think this will introduce type instability.
```suggestion
        push!(get!(d, f(v), Vector{eltype(list)}()), v)
```

src/import_edf.jl
```diff
@@ -204,22 +222,49 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
...
+            edf_channels = collect(edf_channels)
+            info = edf_signals_to_samplesinfo(edf, edf_channels, first(signal_names), collect(edf_channel_names))
+            return info, edf_channels
+        catch e
+            units = [s.header.label => s.header.physical_dimension for s in edf_channels]
+            msg ="""Skipping signal: error while processing units and encodings
+                    for $(first(signal_names)) signal with units $units"""
+            return SamplesInfoError(msg, e)
+        end
+    end
```
kleinschmidt, line 264: I'm having a hard time following what's going on.  Either more informative variable names (what are the pairs? what is grouped?) or some comments would help a lot I think.

src/import_edf.jl
```diff
@@ -299,13 +344,20 @@ following transformations:
...
     file_format = "lpcm.zst"
 
     signals = Any[]
-    edf_samples = edf_to_onda_samples(edf; custom_extractors=custom_extractors)
+    edf_samples, errors = edf_to_onda_samples(edf; custom_extractors=custom_extractors)
+    for error in errors
+        if isa(e, SamplesInfoError)
+            @warn e.msg exception=e.cause
+        else
+            @warn exception=e
```
kleinschmidt, line 358: If it's really critical that these kinds of errors don't actually cause the job to crash/fail then they could use tests. (and same for above with the `SamplesInfoError`)

src/import_edf.jl
```diff
@@ -338,6 +390,30 @@ function validate_arrow_prefix(prefix)
...
+        summary = (info.kind, channel_name, info.sample_unit)
+        push!(ambiguous, (info, edf_signals) => summary)
+    end
+    length(ambiguous) < 2 && return nothing
+    for ((info, edf_signals), (_, channel_name, _)) in ambiguous
+        filter!(!=(channel_name), info.channels)
+        filter!(!=(edf_signal), edf_signals)
+    end
+    return map(last, ambiguous)
+end
```
kleinschmidt, line 416: I haven't thought this through super carefully but I have a suspicion that unless you're very careful there's a risk of removing BOTH+ versions of the ambiguous extracted signal...or  maybe that's what you want?  Either way could you add a comment explaining?

src/standards.jl
```diff
@@ -70,33 +72,43 @@ const STANDARD_LABELS = Dict(# This EEG channel name list is a combined 10/20 an
                                          "tp9", "tp7", "cp5", "cp3", "cp1", "cpz", "cp2", "cp4", "cp6", "tp8", "tp10",
                                          "t5", "p9", "p7", "p5", "p3", "p1", "pz", "p2", "p4", "p6", "p8", "p10", "t6",
                                          "po7", "po3", "poz", "po4", "po8",
-                                         "o1", "oz", "o2",
+                                         "o1" => ["01"], "oz", "o2" => ["02"],
```
kleinschmidt, line 75: god are we really seeing "zero-one" for `o1` in the wild??

kleinschmidt (commented):
This is a monumental achievement in data wrangling (or at least the tip of that iceberg :)

src/import_edf.jl
kolia: Yeah I'm removing all/both extracted channels, because I don't know which to pick.

I thought of checking whether one of the signal type strings ("emg", "eeg" etc) appears anywhere in the label or transducer_type, then pick that one. Not sure how much of the wild data that would cover / whether that's worth the trouble.

src/standards.jl
kolia: 0 and O are right next to each other on the keyboard. From this, we can estimate the diameter of the sleep technicians fingertips.

src/import_edf.jl
kolia: The test that takes in `test/test_edf_to_samples_info.out` is fairly thorough, though it's not systematic, just has a lot of varied test cases...

src/import_edf.jl
kolia: I pushed some comments, hopefully that helps.

I also added a bullet point in the main comment above explaining the grouping.

kolia:
I did a minor refactor to expose diagnostic info to users that were in tests, pushing that here now, since I'm not seeing reviewing activity.

src/import_edf.jl
```diff
@@ -2,32 +2,22 @@
...
+#   otherwise the channel extraction will be rejected.
 function _normalize_references(original_label, canonical_names)
-    label = replace(lowercase(original_label), r"\s"=>"")
+    label = replace(_safe_lowercase(original_label), r"\s"=>"")
     label = replace(replace(label, '('=>""), ')'=>"")
+    label = replace(label, r"\*$"=>"")
     label = replace(label, '-'=>'…')
     label = replace(label, '+'=>"…+…")
     label = replace(label, '/'=>"…/…")
+    label = !isnothing(match(r"^\[.*\]$", label)) ? label[2:end-1] : label
```
ararslan, line 20:
```suggestion
    m = match(r"^\[(.*)\]$", label)
    if m !== nothing
        label = only(match.captures)
    end
```
This is a little easier to follow, IMO.

src/import_edf.jl
```diff
@@ -51,14 +41,21 @@ function _normalize_references(original_label, canonical_names)
...
 function match_edf_label(label, signal_names, channel_name, canonical_names)
-    type, spec = edf_type_and_spec(lowercase(label))
-    if spec === nothing
-        label = type
-    else
-        any(==(type), signal_names) || return nothing
-        label = spec
+    label = _safe_lowercase(label)
+    for signal_name in signal_names
+        m = match(Regex("[\\s\\[,\\]]*$(signal_name)[\\s,\\]]*\\s+(?<spec>.+)", "i"), label)
```
ararslan, line 52: This one heck of a regex... Could you add a comment above it or something giving an example of what you expect it to match?

src/import_edf.jl
```diff
@@ -147,23 +144,21 @@ end
...
 
 function extract_channels(edf_signals, channel_matchers)
-    extracted_channel_names = String[]
-    extracted_channels = EDF.Signal[]
+    extracted_channels = Pair{String,EDF.Signal}[]
     for channel_matcher in channel_matchers
         for edf_signal in edf_signals
             edf_signal isa EDF.Signal || continue
-            any(x -> x === edf_signal, extracted_channels) && continue
+            any(x -> x === edf_signal, map(last, extracted_channels)) && continue
```
ararslan, line 151:
```suggestion
            any(x -> last(x) === edf_signal, extracted_channels) && continue
```

src/import_edf.jl
```diff
@@ -147,23 +144,21 @@ end
...
             channel_name = channel_matcher(edf_signal)
             channel_name === nothing && continue
-            push!(extracted_channel_names, channel_name)
-            push!(extracted_channels, edf_signal)
+            push!(extracted_channels, channel_name => edf_signal)
         end
     end
-    return extracted_channel_names, extracted_channels
+    return extracted_channels
```
ararslan, line 157: We'll have to register this change as a breaking version, right, since the return type is changing? Or is this function internal only?

src/import_edf.jl
```diff
@@ -204,22 +229,53 @@ can be either a `String` giving the generated channel name, or a `Pair` mapping
...
+        # x is either a channel name (string), or a channel_name => alternatives Pair.
+        this_channel_name = x isa Pair ? first(x) : x
+        return s -> match_edf_label(preprocess_labels(s.header.label, s.header.transducer_type),
+                                    signal_names,
+                                    this_channel_name,
                                     channel_names)
     end
-    edf_channel_names, edf_channels = extract_channels(edf.signals, (matcher(x) for x in channel_names))
-    isempty(edf_channel_names) && return nothing
+    edf_channels = extract_channels(edf.signals, (matcher(x) for x in channel_names))
```
ararslan, line 253: Should the `isempty` check remain?

src/import_edf.jl
```diff
@@ -281,31 +339,40 @@ Collections of `EDF.Signal`s are mapped as channels to `Onda.Signal`s via simple
...
+- the label's prepended signal type is matched against known types, if present
+- the remainder of the label is whitespace-stripped, parens-stripped, and lowercased
 - trailing generic EDF references (e.g. "ref", "ref2", etc.) are dropped
 - any instance of `+` is replaced with `_plus_` and `/` with `_over_`
 - all component names are converted to their "canonical names" when possible
   (e.g. for an EOG matched channel, "eogl", "loc", "lefteye", etc. are converted
   to "left").
 
+`EDF.Signal`s which get extracted into more than one `Onda.Samples` are removed
+and a `AmbiguousChannelError` displayed as a warning.
```
ararslan, line 359:
```suggestion
and an `AmbiguousChannelError` displayed as a warning. 
```
Also, this doesn't happen explicitly in this function, but is rather implied by the loop over `diagnostics`, which will make any exception a warning. Should it be documented that every exception becomes a warning rather than calling out this particular type? Or is this the only one that may show up here?

src/import_edf.jl
```diff
@@ -257,7 +313,9 @@ end
 
 """
     store_edf_as_onda(path, edf::EDF.File, uuid::UUID=uuid4();
-                      custom_extractors=(), import_annotations::Bool=true,
+                      custom_extractors=STANDARD_EXTRACTORS, import_annotations::Bool=true,
+                      postprocess_samples=identity,
+                      arrow_folder=nothing,
```
ararslan, line 318:
```suggestion
```
Seems that wasn't actually added?

src/import_edf.jl
```diff
@@ -365,17 +460,69 @@ following transformations:
...
+            ambiguous_channels = _ambiguous_channels!(matched, edf_signal)
+            if !isnothing(ambiguous_channels)
+                edf_signal_summary = (edf_signal.header.label,
+                                      edf_signal.header.transducer_type,
+                                      edf_signal.header.physical_dimension)
+                push!(errors, AmbiguousChannelError(edf_signal_summary => ambiguous_channels))
+            end
+        end
+        matched = [(info, edf_signals) for (info, edf_signals) in matched if !isempty(info.channels)]
+        errors = sort(errors; by=string)
```
ararslan, line 515:
```suggestion
        sort!(errors; by=string)
```

src/import_edf.jl
```diff
@@ -365,17 +460,69 @@ following transformations:
...
+        matched, errors
+    catch e
+        [], [e]
     end
-    return edf_samples
+    matched_edf_headers = reduce(∪, last.(info_map); init=[])
```
ararslan, line 520:
```suggestion
    matched_edf_headers = mapreduce(last, union, info_map; init=[])
```

src/prettyprint_diagnostic_info.jl
```diff
@@ -0,0 +1,101 @@
+function _print_result_value(f, si::SamplesInfo; indent=0, kwargs...)
+    print(f, "$(repeat(' ', indent))SamplesInfo(")
+    print(f, join([repr(getproperty(si, field)) for field in fieldnames(SamplesInfo)], ", "))
```
ararslan, line 3:
```suggestion
    join(f, (repr(getfield(si, field)) for field in fieldnames(SamplesInfo)), ", "))
```

src/prettyprint_diagnostic_info.jl
```diff
@@ -0,0 +1,101 @@
...
+    print(f, ")")
+end
+
+function _print_result_value(f, p::Pair{A, Vector{B}}; indent=0, tab=4) where {A, B}
```
ararslan, line 7:
```suggestion
function _print_result_value(f, p::Pair{A,Vector{B}}; indent=0, tab=4) where {A,B}
```

src/prettyprint_diagnostic_info.jl
```diff
@@ -0,0 +1,101 @@
...
+    _print_result_value(f, first(p); indent=indent, tab=tab)
+    println(f, " => [")
+    for x in last(p)
+        _print_result_value(f, x; indent=indent+tab, tab=tab)
```
ararslan, line 11:
```suggestion
        _print_result_value(f, x; indent=(indent + tab), tab=tab)
```

src/prettyprint_diagnostic_info.jl
```diff
@@ -0,0 +1,101 @@
...
+        print_result(f, result, i; indent=4)
+        println(f, ",")
+    end
+    print(f, "]")
+end
+
+_edf_headers(r) = vcat(map(last, r.header_map)..., r.unextracted_edf_headers)
+
+_groupby(r) = Set((h.label, h.transducer_type, h.physical_dimension)
+                 for h in _edf_headers(r))
```
ararslan, line 50:
```suggestion
                  for h in _edf_headers(r))
```

src/prettyprint_diagnostic_info.jl
```diff
@@ -0,0 +1,101 @@
...
+mock_edf_signal(header, n_records) = EDF.Signal(header, rand(Int16, n_records * header.samples_per_record))
+
+"""
+    mock_edf(diagnostic_info)
+
+Turn a diagnostic info `NamedTuple` as returned by `edf_to_onda_samples` or 
+`edf_header_to_onda_samples_info` and contained in `test_edf_to_samples_info.in`
+into a mock edf, for testing purposes.
+"""
+function mock_edf(result)
```
ararslan, line 91: Seems like the mocking stuff should be part of the test suite rather than the package code.

src/standards.jl
```diff
@@ -70,33 +72,43 @@ const STANDARD_LABELS = Dict(# This EEG channel name list is a combined 10/20 an
...
+                                         "right_anterior_tibialis"=> ["rat", "rat1", "r", "right", "rightlimb", "tibr", "tibre", "plmr", "leg2", "leg3", "rleg", "rleg1", "legr", "jambe_r", "rightleg"]],
+                             # it is common to see ambiguous channels, which could be leg or face EMG
+                             # if leg EMG is present in separate channels,
+                             # post-processing might map "emg_ambiguous"
+                             # to chin channels (for example)
+                             ["emg_ambiguous", "emg"] => ["1" => ["aux1", "l"],
+                                                          "2" => ["aux2", "r"],
+                                                          "3" => ["aux3"]],
+                             ["heart_rate"] => ["heart_rate"=> ["hr", "pulse", "pulso", "pr", "pulserate"]],
+                             ["snore"] => ["snore" => ["ronquido", "ronquido derivad", "schnarchen", "ronfl", "schnarchmikro"]],
```
ararslan, line 102: Ronquido, schnarchen, shnarchmikro... I'm going to bet these are some variant of "snore" in non-English languages. IMO other languages should be left for custom extractors, and we should probably document that we cover common English names only.

src/import_edf.jl
kolia: ooh `only`

src/prettyprint_diagnostic_info.jl
kolia: Yeah I initially had it in the test suite, but moved it here because I wanted to use it from downstream (from the actual ingest code.)

The use-case is that when you're ingesting a large set of EDFs, this mock tooling is a convenient way of working.

(I'm assuming downstream packages can't depend on the functions defined in the test suite somehow, can they?)

src/standards.jl
kolia: My policy here has been to include alternative spellings / languages if they are un-ambiguous.

src/prettyprint_diagnostic_info.jl
ararslan: > (I'm assuming downstream packages can't depend on the functions defined in the test suite somehow, can they?)

Correct, they cannot.

It's a sufficiently small piece of code that IMO it could just be copied to where ingest is happening, with it living here in the test suite.

ararslan:
Welp, looks like one of the changes broke CI... (also note that there are more review comments that GitHub auto-collapsed)

src/import_edf.jl
kolia, line 55:
```suggestion
        # match exact STANDARD (or custom) signal types at beginning of label, ignoring case
        # possibly bracketed by or prepended with `[`, `]`, `,` or whitespace
        # everything after is included in the spec a.k.a. label
        m = match(Regex("[\\s\\[,\\]]*$(signal_name)[\\s,\\]]*\\s+(?<spec>.+)", "i"), label)
```

src/import_edf.jl
kolia: I'd say `extract_channels_by_label` is the public-facing entrypoint, and `extract_channels` is internal.

src/import_edf.jl
kolia: Could add a similar bypass back in like so

```
isempty(edf_channels) && return ([], [])
```

but it's not necessary, since that's what already gets returned without the bypass.

src/import_edf.jl
kolia:
```suggestion
and any errors captured during channel extraction displayed as warnings, principally
- `AmbiguousChannelError`s indicating that the same EDF signal can be extracted into
more than one signal type and will hence be ignored
- `SampleInfosError` indicating an error on constructing a `SamplesInfo`,
most commonly because sample units do not match
any known standard physical unit. 
```

kleinschmidt:
Need to fix failing CI before this can be merged

kleinschmidt:
Hey I fixed CI without even really meaning to! :tada:

kolia:
Whoa does this mean we can actually merge this?

kleinschmidt (approved):

PR #27, opened 2021-09-23
bump onda compat to add 0.14

kleinschmidt:
Minimal changes, just had to add a polyfill for `Onda.samples_info` for Onda < v0.14 (tested locally and works on Onda 0.13 at least)


kolia (approved):
yolo

PR #29, opened 2021-10-26
Use EDF 0.7

omus:
If need be we can support both EDF 0.6 and 0.7 at the same time

### DIFF
--- a/src/export_edf.jl
+++ b/src/export_edf.jl
@@ -97,7 +97,7 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
 end
 
 function onda_samples_to_edf_signals(onda_samples::AbstractVector{<:Samples}, seconds_per_record::Float64)
-    edf_signals = Union{EDF.AnnotationsSignal,EDF.Signal}[]
+    edf_signals = Union{EDF.AnnotationsSignal,EDF.Signal{Int16}}[]
     for samples in onda_samples
         # encode samples, rescaling if necessary
         if sizeof(sample_type(samples.info)) > sizeof(Int16)
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Compat = "3.32"
-EDF = "0.6"
+EDF = "0.7"
 FilePathsBase = "0.9"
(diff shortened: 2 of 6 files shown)


src/export_edf.jl
```diff
@@ -97,7 +97,7 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
 end
 
 function onda_samples_to_edf_signals(onda_samples::AbstractVector{<:Samples}, seconds_per_record::Float64)
-    edf_signals = Union{EDF.AnnotationsSignal,EDF.Signal}[]
+    edf_signals = Union{EDF.AnnotationsSignal,EDF.Signal{Int16}}[]
```
jrevels, line 100: What's the benefit to adding this parameter restriction? I could be missing something, but seems like this is an implementation detail and I don't see a correctness/performance benefit to surface/codify it?

for example, it feels like this code shouldn't break if the `EDF.Signal` constructor decided to e.g. promote provided encoded sample data to Int32 (not that that would happen, just kinda making a hypothetical point)

ararslan: This `Union` is the type of the `signals` field in an `EDF.File` object now that `EDF.Signal` is parameterized on the encoded sample type (for BDF support in EDF.jl v0.7). I think the purpose of adding it here is that you won't be able to construct an `EDF.File` programmatically with a `Union{EDF.AnnotationsSignal,EDF.Signal}`, since `convert` that's called automatically with inner constructors will fail.

kolia (commented):
If I understand the underlying change to EDF, this should allow OndaEDF to read in BDF without any code changes, and the changes in this PR are so that when you export it'll always write EDF (?)

Should there be a test that we can actually read BDF into Onda?

omus:
> Should there be a test that we can actually read BDF into Onda?

The scope of this change it just to have an OndaEDF.jl version which is compatible with the latest EDF.jl version.

kolia (approved):

omus:
Rebased to address conflict and I also bumped the version to 0.9.2

jrevels (changes requested):
Before merging, this requires https://github.com/beacon-biosignals/OndaEDF.jl/pull/35 to be merged into it to fix the problem uncovered in #34 

(also that needs a test)

jrevels (approved):

jrevels:
Approved now that #35 is merged into this PR. will merge assuming tests pass

Issue #34, opened 2021-11-26
`edf_to_onda_samples` didn't return correct sample data as expected


likanzhan:
In the [Dropbox Folder](https://www.dropbox.com/sh/kr04xkj4vjsubqn/AAAv9urE9VI5F35JYLqquLsfa?dl=0), I uploaded a sample data recorded in our study. We have a data file `data.bdf` and a trigger data file `evt.bdf`. Normally the data is processed with Matlab, `readbdfdata.m`.

For now, we can use the `EDF.read()` to read the data, and the `EDF.decode`d data are similar to the results obtained with MatLab (appendix 1)

```julia
edf = EDF.read("data.bdf")
dt = [transpose(EDF.decode(signal)) for signal in edf.signals]
data_edf = reduce(vcat, dt)
```

But the output of the `edf_to_onda_samples()` function is totally different from that of the supposed results (appendix 2):

```julia
edf_samples, nt = edf_to_onda_samples(edf)
data_onda = edf_samples[1].data

```

What is the problem here? 

Appendixes:

- `data_edf`:

```
64×2034000 Matrix{Float32}:
  -4040.22        -4041.19         -4043.19       …    -6332.75         -6334.94        -6333.69
 -14271.8        -14270.7         -14272.1            -16735.0         -16743.6        -16743.9
 -10168.2        -10167.1         -10168.6            -11401.8         -11400.3        -11397.1
 -20942.8        -20941.9         -20943.9            -21258.4         -21259.2        -21257.4
 -14660.4        -14661.4         -14663.3            -14604.1         -14602.7        -14600.2
 -18451.6        -18448.0         -18448.7        …   -21246.9         -21248.2        -21245.2
   -709.656        -709.062         -709.594           -1886.06         -1882.62        -1879.66
  10168.4         10168.6          10168.4             10640.8          10641.2         10642.9
  -5651.0         -5649.78         -5649.19            -5908.84         -5907.38        -5905.12
  -5041.62        -5041.81         -5042.62            -5764.62         -5763.41        -5762.16
  -3191.5         -3190.28         -3189.12       …    -6197.94         -6197.0         -6193.91
  -5846.34        -5846.44         -5846.25            -4642.69         -4638.91        -4638.0
 -20708.8        -20708.8         -20708.5            -20613.6         -20611.0        -20604.5
  -6359.38        -6358.88         -6358.78           -10465.7         -10462.1        -10462.3
[...]

PR #35, opened 2021-11-26
support 24-bit sample types (BDF)

jrevels:
ref #34

still needs tests + a more current maintainer should check behind me on this one

### DIFF
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -107,14 +107,15 @@ function edf_signal_encoding(edf_signal_header::EDF.SignalHeader,
     sample_resolution_in_unit = (pmax - pmin) / (dmax - dmin)
     sample_offset_in_unit = pmin - (sample_resolution_in_unit * dmin)
     sample_rate = edf_signal_header.samples_per_record / edf_seconds_per_record
+    sample_type = (dmax > typemax(Int16) || dmin < typemin(Int16)) ? Int32 : Int16
     return (sample_resolution_in_unit=Float64(sample_resolution_in_unit),
             sample_offset_in_unit=Float64(sample_offset_in_unit),
-            sample_rate=Float64(sample_rate))
+            sample_rate=Float64(sample_rate),
+            sample_type=sample_type)
 end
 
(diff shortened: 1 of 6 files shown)


jrevels:
I've updated this PR to be against #29, since it resolves a bug with that PR

src/import_edf.jl
```diff
@@ -107,14 +107,15 @@ function edf_signal_encoding(edf_signal_header::EDF.SignalHeader,
...
             sample_offset_in_unit=Float64(sample_offset_in_unit),
-            sample_rate=Float64(sample_rate))
+            sample_rate=Float64(sample_rate),
+            sample_type=sample_type)
 end
 
-# TODO: implement a more disciplined widening protocol for `sample_type`
 function promote_encodings(encodings; pick_offset=(_ -> 0.0), pick_resolution=minimum)
-    sample_type = Int16
+    sample_type = reduce(promote_type, (e.sample_type for e in encodings))
```
ararslan, line 118:
```suggestion
    sample_type = mapreduce(e -> e.sample_type, promote_type, encodings)
```

jrevels: AFAICT this is subjective, but I prefer the generator approach/find it more legible

omus, line 118: I'll call out that `promote_type` takes varargs:
```suggestion
    sample_type = promote_type((e.sample_type for e in encodings)...)
```

omus (approved):

omus:
> still needs tests

Let's just make an issue for this so these changes to stay in limbo forever

jrevels:
> Let's just make an issue for this so these changes don't stay in limbo forever

a basic test should be simple, i'll add one today

src/import_edf.jl
jrevels: Yeah, but I prefer not to splat unless i'm in a situation where the splatted element count is known locally (here it's not - it's caller driven)

jrevels:
given the approval i'll go ahead and merge this in

Issue #40, opened 2022-01-11
Directory creation failure with `S3Path` and `store_edf_as_onda`


omus:
When passing in an `S3Path` as the `onda_dir` into `store_edf_as_onda` an error will occur if the `samples` subdirectory does not exist:
```julia
ERROR: ArgumentError: S3Path folders must end with '/': s3://bucket/prefix/file.edf/samples
Stacktrace:
  [1] mkdir(fp::AWSS3.S3Path{Nothing}; recursive::Bool, exist_ok::Bool)
    @ AWSS3 ~/.julia/packages/AWSS3/bwC3m/src/s3path.jl:401
  [2] mkpath
    @ ~/.julia/packages/FilePathsBase/kMipS/src/aliases.jl:17 [inlined]
  [3] store_edf_as_onda(edf::EDF.File{IOBuffer}, onda_dir::AWSS3.S3Path{Nothing}, recording_uuid::Base.UUID; custom_extractors::Vector{OndaEDF.var"#2#4"{_A, Vector{String}} where _A}, import_annotations::Bool, postprocess_samples::typeof(TemporarySpikeRunner.correct_channel_names!), signals_prefix::String, annotations_prefix::String)
    @ OndaEDF ~/.julia/packages/OndaEDF/Z5Aq7/src/import_edf.jl:380
...
```
In this example `p"s3://bucket/prefix/file.edf/"` was passed in. The creation of the subdirectory is handled by `store_edf_as_onda` and we should add `* '/`` when creating the samples dir to have this code work generically.

omus:
The work around to this problem is to pre-create the samples subdir prior to calling `store_edf_as_onda`

jrevels:
if you're using S3 it might also be good to not use the high-level convenience functions (meant more for scripts/interactive work) but instead use the more primitive API functions in composition with the save/load functions you want, e.g.

https://beacon-biosignals.github.io/OndaEDF.jl/stable/#OndaEDF.edf_to_onda_samples

IIRC i leaned toward getting rid of `store_edf_as_onda` because was afraid people would want to keep adding features to it instead of more simply composing with other functions 😁

omus:
> if you're using S3 it might also be good to not use the high-level convenience functions

Overall I agree with the sentiment of using lower-level functions and making use of composition. As I'm trying to quickly iterate through this work I'll make the fix for this here and look into refactoring my code as a follow up.

PR #41, opened 2022-01-11
Make `store_edf_as_onda` compatible with `S3Path`

omus:
Fixes: https://github.com/beacon-biosignals/OndaEDF.jl/issues/40


kleinschmidt (approved):
I think this is probably fine.  Are there any conceivable circumstances when having an extra trailing slash would cause a problem?

omus:
> Are there any conceivable circumstances when having an extra trailing slash would cause a problem?

None that I can think of. Even if we end up with multiple ending trailing slashes `mkpath` should work fine

Issue #36, opened 2021-11-26
emit unscaled floating-point-encoded signals instead of re-encoding when input encodings are mismatched


jrevels:
realized while writing https://github.com/beacon-biosignals/OndaEDF.jl/issues/34#issuecomment-980097812

Let's say OndaEDF extracts a group of channels to convert into a single `Onda.Samples`. If all channels have the same LPCM encoding, OndaEDF will just use that, concatenate the channels into a matrix, and boom, done. If the channels have different LPCM encodings, however, OndaEDF will try to select the "best" target encoding and re-encode all of the channels before concatenation. This re-encoding process involves dithering, which is kind of computationally expensive (requires RNG) and very slightly raises the signal's noise floor (only by the minimal step size of the encoding resolution - well worth it to avoid the quantization distortion that could occur otherwise).

OndaEDF does this because when OndaEDF was originally written, Onda didn't have floating-point encodings, so OndaEDF HAD to select some integer encoding. These days, though, Onda supports floating-point encodings.

Thus, I think we should just remove the whole "encoding selection" machinery entirely if the channels don't all have the same LPCM encoding. Instead, we should just emit an unscaled floating-point encoding (e.g. `sample_resolution_in_unit=1.0`, `sample_offset=0.0`, `sample_type=Float32`). That way no dithering is required and we can delete a bunch of code

jrevels:
i'd imagine for most EDFs the DC offset will be low enough and/or the actual sensor precision will be low enough that floating point precision wouldn't come into play 

could be wrong though

PR #45, opened 2022-02-15
try to make import experience better

kleinschmidt:
The basic idea is that there are two stages: 

1. First you _plan_ the conversion, which yields a table/dataframe with one row per EDF signal and the samplesinfo that will be generated from that signal alone.  You can inspect this, save it out as a record, manipulate to adjust the plan as necessary, and then...
2. You _execute_ the plan, converting groups of EDF signals defined by groups of rows of the plan table into Onda.Samples objects.

The motivation for this is to make it easier to have a persistent record of how an EDF file (or collection of them) was transformed into Onda, which can be easily queried to see which channels/signals didn't make it through.  It also aims to reduce the indirection in the code that implements the channel/signal extraction logic to make it easier to understand what's going on there and easier to see how to create custom extractors.  It doesn't go as far as maybe it could there but it's a start (by inverting some loops to operate on a single signal at a time and removing the frankly tortured closure-in-a-closure design)

Remaining things to do/decide:
- update: yes, I've implemented this now.  should `plan(edf::EDF.File)` do the `promote_encoding` stuff that's done in `execute_plan` currently?  ~~it doesn't now, but~~ I'm thinking it might be better to do it that way so that you catch any potential issues there
- is `plan` too vague as a function name?  maybe it should be `make_plan` or something more unambiguously verby
- should we be uniqueifying `kind` here, or leave that up to the user?  it's easy enough to insert some middle-ware that modifies the plan but a potential gotcha in situations where you're dealing with systems that assume that `kind` is unique per recording.
- update: yes, I did this.  ~~should we define a legolas schema for the plan table?  it's easy enough to do but I wasn't sure we wanted to take the extra dependency (although it's already indirectly included because of Onda I guess)~~

a few other misc (and potentially breaking) changes:
- while profiling I discovered that compiling the regex for the signal name stripping was a HUGE bottleneck (~90% of planning time).  I replaced it with something that's usually equivalent but will not be in cases where the `signal_name` contains a whitespace or another delimiter character `,[]`.  none of the test cases are like this so I figured it was okay, and it gives you a free 10x speedup (planning one EDF.File went from 80ms to 8ms)
- the use of "custom extractors" has been removed in favor of directly passing in labels of the form `["canonical_kind", "other_kinds"...] => ["channel1", "channel2_canonical" => ["channel2_alt"...], ...]`.  I think this is justified because in practice nearly all uses of custom extractors were using custom labels, rather than non-label-based functions.  and the plan table provides a useful intermediate representation that can be manipulated in a way that is much more flexible and less restrictive than the custom extractors provided.
- the order of the extracted channels is now determined by their order in the EDF file, rather than their order in the labels (tests have changed to reflect this)
- lastly, I've removed all the code for pretty-printing the diagnostics and banished the glue code I'd originally written to convert the old diagnostics to/from the plan tables to `test/diagnostics/`.  I didn't find the diff with with the existing diagnostics file to be useful since there's some non-determinism to the order things get printed because the header map used Dicts, and their iteration order is undefined.  and the notion of the plan table replaces these diagnostics and is easier to work with programmatically (it's just a table).  When developing this the change in the regex for matching signal names in labels made a handful of these tests not pass for reasons that were...mysterious to me, but it was close enough that I wasn't concerned.  I think the better visibility that the plan provides and the massive speed up justifies a bit of breakage there.
- missing units no longer cause an error; instead you get a missing value in `sample_unit`. I have no idea why that was done in that way before.


src/import_edf.jl
```diff
@@ -51,16 +96,11 @@ _safe_lowercase(s::AbstractString) = map(_safe_lowercase, s)
...
-    for signal_name in signal_names
-        # match exact STANDARD (or custom) signal types at beginning of label, ignoring case
-        # possibly bracketed by or prepended with `[`, `]`, `,` or whitespace
-        # everything after is included in the spec a.k.a. label
-        m = match(Regex("[\\s\\[,\\]]*$(signal_name)[\\s,\\]]*\\s+(?<spec>.+)", "i"), label)
-        if !isnothing(m)
-            label = m[:spec]
-        end
-        # if signal type does not match, use the entire label
+    m = match(r"[\s\[,\]]*(?<signal>.+?)[\s,\]]*\s+(?<spec>.+)"i, label)
```
palday, line 99: add some breadcrumbs here please

src/import_edf.jl
```diff
@@ -100,8 +140,7 @@ end
 #   pmax + pmin + (pmax*(dmin - dmax)/(dmax - dmin)) ≈ pmin
 #   pmax + pmin + (-pmax) ≈ pmin
 #   pmin ≈ pmin
-function edf_signal_encoding(edf_signal_header::EDF.SignalHeader,
-                             edf_seconds_per_record)
+function edf_signal_encoding(edf_signal_header, edf_seconds_per_record)
```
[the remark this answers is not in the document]
kleinschmidt: because this could be a (legolas style) row with extra fields.  all that matters here is that it has the necessary properties

src/import_edf.jl
```diff
@@ -49,18 +94,75 @@ _safe_lowercase(c::Char) = isvalid(c) ? lowercase(c) : c
 # malformed UTF-8 chars are a choking hazard
 _safe_lowercase(s::AbstractString) = map(_safe_lowercase, s)
 
+"""
+    match_edf_label(label, signal_names, channel_name, canonical_names)
+
+Return a normalized label matched from and EDF `label`.  The purpose of this
```
alexmchan, line 100:
```suggestion
Return a normalized label matched from an EDF `label`.  The purpose of this
```

alexmchan:
One specific issue that comes up in the [MESA data](https://github.com/beacon-biosignals/data-map/issues/380) is that the `-` character is assumed to be involved with indicating EEG referencing, but in the MESA case, it's used to separate out the signal from the channel name (e.g. `EOG-L` and `EOG-R`). This behavior doesn't look like it changes with this PR, but it seems that what would now be required is manipulating the plan table before executing (rather than building custom extractors). Doesn't seem like that's an issue, but I'm wondering if there's a way to address the types of naming conventions that break the built-in assumptions before it gets to doing plan manipulation.

kleinschmidt:
Yeah that's the kind of thing I'd handle with a custom label(s); the goal here is to make it easier to diagnose/iterate on that kind of issue.

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
+# "channel" => ["alt1", "alt2", ...]
+canonical_channel_name(channel_alternates::Pair) = first(channel_alternates)
 
-For one or more signal names and one or more channel names,
-return a list of `[(infos, errors)...]` where `errors` is a vector of errors that occurred,
-and `infos` is a vector of `(si::Onda.SamplesInfo, edf_signals::Vector{EDF.Signal})`,
-where `edf_signals` align with `si.channel`. This list can have more than one pair
-if channels with the same signal kind/type have different sample rates or
-physical units.
+plan_edf_to_onda_samples(signal::EDF.Signal, s; kwargs...) = plan_edf_to_onda_samples(_named_tuple(signal.header), s; kwargs...)
```
ararslan, line 311: Doesn't actually matter (feel free to ignore) but I'd have this method call the other, since you're saying that the way to deal with a signal is to treat it like its header.
```suggestion
plan_edf_to_onda_samples(signal::EDF.Signal, s; kwargs...) = plan_edf_to_onda_samples(signal.header, s; kwargs...)
```

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
+plan_edf_to_onda_samples(header::EDF.SignalHeader, s; kwargs...) = plan_edf_to_onda_samples(_named_tuple(header), s; kwargs...)
 
-`(label, transducer_type)` pairs will be transformed into labels by `preprocess_labels`
-(default preprocessor returns the original label). An alternate to `STANDARD_UNITS`
-for mapping different spellings of units to their canonical unit name can be passed in
-as `unit_alternatives`.
+"""
+    plan_edf_to_onda_samples(header, seconds_per_record; labels=STANDARD_LABELS,
+                             units=STANDARD_UNITS, preprocess_labels=(l,t) -> l)
```
ararslan, line 316: Looks like the purpose of the `preprocess_labels` keyword argument isn't documented. I don't know what I expected but I did not expect the arguments passed to it to be the label and the transducer type as I've never seen the transducer type used for anything meaningful before. Since `preprocess_labels` is a function of `header` only, does it have to be the job of `plan_edf_to_onda_samples` to call it?

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
+                   "avl"=> ["ecgl", "ekgl", "ecg", "ekg", "l"], 
+                   "avr"=> ["ekgr", "ecgr", "r"], ...]
+```
+
+Matching is done in the order that `labels` iterates pairs, and will stop at the
+first match, with no warning if signals are ambiguous (although this may change
+in a future version)
+"""
+function plan_edf_to_onda_samples(header,
+                                  seconds_per_record=_get(header,
```
ararslan, line 361: Is it ever the case that you'd want to call this function with some arbitrary value of `seconds_per_record` that doesn't necessarily match what was in the file header? If not, you can move your `ismissing` check below the `row` definition and use `ismissing(_get(row, :seconds_per_record))`. It will have it if `header` has it.

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
+                matched = match_edf_label(edf_label, signal_names, channel_name, channel_names)
+                
+                if matched !== nothing
+                    # create SamplesInfo and return
+                    row = rowmerge(row; 
+                                   channel=matched,
+                                   kind=first(signal_names),
+                                   sample_unit=edf_to_onda_unit(header.physical_dimension, units),
+                                   edf_signal_encoding(header, seconds_per_record)...,
+                                   )
```
ararslan, line 389:
```suggestion
                                   edf_signal_encoding(header, seconds_per_record)...)
```

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
+                                  units=STANDARD_UNITS,
+                                  preprocess_labels=(l,t) -> l,
+                                  onda_signal_groups=grouper((:kind, :sample_unit, :sample_rate)))
+    # remove non-Signals (e.g., AnnotationsSignals), keeping track of indices
+    enum_signals = [(i, s) for (i, s) in enumerate(edf.signals) if s isa EDF.Signal]
+    plan_rows = map(enum_signals) do (edf_signal_idx, signal)
+        row = plan_edf_to_onda_samples(signal.header, edf.header.seconds_per_record;
+                   labels, units, preprocess_labels)
+        return rowmerge(row; edf_signal_idx)
+    end
```
ararslan, line 438: This was a bit difficult to follow at first. I almost feel like it'd be more straightforward to just to it all in one go:
```suggestion
    plan_rows = [rowmerge(plan_edf_to_onda_samples(s.header, edf.header.seconds_per_record;
                                                   labels, units, preprocess_labels);
                          edf_signal_idx=i)
                 for (i, s) in enumerate(edf.signals)
                 if s isa EDF.Signal]
```

src/import_edf.jl
```diff
@@ -208,81 +302,264 @@ function groupby(f, list)
...
-    end
+Convert Signals found in an EDF File to `Onda.Samples` according to the plan
+specified in `plan_table` (e.g., as generated by [`plan_edf_to_onda_samples`](@ref)), returning an
+iterable of the generated `Onda.Samples` and the plan as actually executed.
 
-    results = map(values(grouped)) do pairs
-        # pairs is a vector of `standard_onda_name::String => EDF.Signal` pairs
-        edf_channel_names, edf_channels = zip(pairs...)
+The input plan is transformed by using [`merge_samples_info`](@ref) to combine
+rows with the same `:onda_signal_idx` (or output of `sample_groups`) into a
```
ararslan, line 466: What's an example of a case where you might want to group by something other than the Onda signal index, i.e. the grouping already determined in `plan_table`? IMO this seems like a weird thing to support but perhaps I'm not seeing its  f u l l   p o t e n t i a l.

src/import_edf.jl
```diff
@@ -331,45 +608,37 @@ in `\$path/samples/`, and write the Onda signals and annotations tables to
...
+- the label is whitespace-stripped, parens-stripped, and lowercased
 - trailing generic EDF references (e.g. "ref", "ref2", etc.) are dropped
 - any instance of `+` is replaced with `_plus_` and `/` with `_over_`
 - all component names are converted to their "canonical names" when possible
-  (e.g. for an EOG matched channel, "eogl", "loc", "lefteye", etc. are converted
-  to "left").
-
-`EDF.Signal`s which get extracted into more than one `Onda.Samples` are removed
-and an `AmbiguousChannelError` displayed as a warning.
+  (e.g. "m1" in an EEG-matched channel name will be converted to "a1").
```
ararslan, line 633: Is that true? I thought we explicitly removed that conversion a while back.

src/import_edf.jl
```diff
@@ -382,10 +651,18 @@ function store_edf_as_onda(edf::EDF.File, onda_dir, recording_uuid::UUID=uuid4()
     mkpath(joinpath(onda_dir, "samples") * '/')
 
     signals = Onda.Signal[]
-    edf_samples, diagnostics = edf_to_onda_samples(edf; custom_extractors=custom_extractors)
-    for e in diagnostics.errors
-        @warn sprint(showerror, e)
+    edf_samples, plan = edf_to_onda_samples(edf; kwargs...)
+    
+    errors = _get(Tables.columns(plan), :errors)
```
ararslan, line 656: As it stands, this will always be `missing` because the column is actually called `error`. (I had to go back and recheck haha)
```suggestion
    errors = _get(Tables.columns(plan), :error)
```

src/import_edf.jl
```diff
@@ -407,17 +684,7 @@ function store_edf_as_onda(edf::EDF.File, onda_dir, recording_uuid::UUID=uuid4()
...
-
-function store_edf_as_onda(path, edf::EDF.File, uuid::UUID=uuid4(); kwargs...)
-    Base.depwarn("`store_edf_as_onda(path, edf, ...)` is deprecated, use " *
-                 "`nt = store_edf_as_onda(edf, path, ...); nt.recording_uuid => (nt.signals, nt.annotations)` " *
-                 "instead.",
-                 :store_edf_as_onda)
-    nt = store_edf_as_onda(edf, path, uuid; kwargs...)
-    signals = [rowmerge(s; file_path=string(s.file_path)) for s in nt.signals]
-    return nt.recording_uuid => (nt.signals, nt.annotations)
+    return @compat (; recording_uuid, signals, annotations, signals_path, annotations_path, plan)
```
ararslan, line 687:
```suggestion
    return (; recording_uuid, signals, annotations, signals_path, annotations_path, plan)
```

ararslan (commented):
I really like this approach! I think this will be a huge improvement.

src/import_edf.jl
kleinschmidt: Yeah, I can imagine a situation where someone might want to split two things that are grouped together into separate groups but not re-compute the indices.  While that might be better handled by custom labels, I don't really see a downside to `reverting control to the caller` here

README.md
```diff
@@ -18,3 +18,28 @@ These expectations are as follows:
...
+    samples, exec_plan = edf_to_onda_samples(edf, signals_plan)
+    return samples, exec_plan
+end
+```
+The executed plan as returned is a Tables.jl compatible table, with one row per EDF.Signal and columns for
+- the fields of the original `EDF.SignalHeader`
+- the fields of the generated `Onda.SamplesInfo`, including
+  - `:kind`, the extracted signal kind
+  - `:channel`, the extracted channel label (instead of `:channels`, since each `EDF.Signal` is exactly one channel in `Onda.Samples`)
+- `:edf_signal_idx`, the numerical index of the source signal in `edf.signals`
```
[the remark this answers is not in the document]
kleinschmidt: yeah good point, and since it's used mostly internally there's really no reason to not spell it out.

src/OndaEDF.jl
```diff
@@ -1,26 +1,131 @@
...
+    Plan = @row("ondaedf-plan@1",
+                # EDF.SignalHeader fields
+                label::String
+                transducer_type::String
+                physical_dimension::String
+                physical_minimum::Float32
+                physical_maximum::Float32
+                digital_minimum::Float32
+                digital_maximum::Float32
+                prefilter::AbstractString
```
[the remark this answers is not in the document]
kleinschmidt: I do think I copy-pasted this directly from the signal hearder struct def, but on double checking it is String...

src/OndaEDF.jl
```diff
@@ -1,26 +1,131 @@
...
+context about the conversion that only apply at the level of an `EDF.File`: 
+- `edf_signal_idx` gives the index of the source EDF signal
+- `onda_signal_idx` gives the index of the output `Onda.Samples`.  Note some
+  indices may be skipped in the output, so this is only to indicate order and
+  grouping.
+
+"""
+const Plan = @row("ondaedf-plan@1",
+                  # EDF.SignalHeader fields
+                  label::String = convert(String, label),
```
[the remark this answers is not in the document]
kleinschmidt: I honestly do not know.  I guess it's conceivable that some other kidn of abstractstring comes in here that needs conversion, or (more likely) some of the fields that are `Float32` come in as other types, but I'm not sure.  I guess it can't hurt?

src/import_edf.jl
```diff
@@ -1,3 +1,71 @@
...
+# XXX: make this @generated for speed:
+function _named_tuple(x::T) where {T}
+    fields = fieldnames(T)
+    values = getfield.(Ref(x), fields)
+    return NamedTuple{fields}(values)
+end
+
+function _err_msg(e, msg="Error while converting EDF:")
+    bt = catch_backtrace()
+    msg = let io = IOBuffer()
```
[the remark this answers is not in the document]
kleinschmidt: not necessary, just makes the grouping clear.  probably a holdover from when I was hacking on this in the repl and didn't want to leave `io` sitting around

src/import_edf.jl
```diff
@@ -1,3 +1,71 @@
...
+
+function _errored_row(row, e)
+    msg = _err_msg(e, "Skipping signal $(row.label): error while extracting channels")
+    return rowmerge(row; error=e)
+end
+
+function _errored_rows(rows, e)
+    labels = [row.label for row in rows]
+    labels_str = join(string.('"', labels, '"'), ", ", ", and ")
+    msg = _err_msg(e, "Skipping signals $(labels_str): error while extracting channels")
```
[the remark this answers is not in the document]
kleinschmidt: actually, we may want to ONLY store the message; Arrow has a hard time serializing the exceptions.

src/import_edf.jl
```diff
@@ -49,18 +117,75 @@ _safe_lowercase(c::Char) = isvalid(c) ? lowercase(c) : c
...
+This returns `nothing` if `channel_name` does not match after normalization
+
+Canonicalization
+
+- ensures the given label is whitespace-stripped, lowercase, and parens-free
+- strips trailing generic EDF references (e.g. "ref", "ref2", etc.)
+- replaces all references with the appropriate name as specified by
+  `canonical_names`
+- replaces `+` with `_plus_` and `/` with `_over_`
+- returns the initial reference name (w/o prefix sign, if present) and the
```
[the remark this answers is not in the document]
kleinschmidt: This is a straight copy paste from the comment above the canonicalization bit above.  I think it refers to the possibility that there's like "EMG CHIN1"

src/import_edf.jl
kleinschmidt: This has been very useful in some data we've seen "in the wild" where the label and transducer were basically flipped in some recordings.  In practice it was much more effective to do this kind of preprocessing rather than try to mutate the EDF file headers.

I do wonder whether, in this brave new world, that kind of situation would be better handled by working directly with the `plan_edf_to_onda_samples(signal::EDF.Signal)` level, since we can :just: manipulate the named tuple for the signal header.  There's some disadvantage to that, sicne the `EDF.File` function does a fair bit of work to sort/group things, which isn't factored out into a separate function that you could apply to a vector of plans.  But maybe this is a good reason to do that kind of refactoring?

src/import_edf.jl
kleinschmidt: hoo boy, good catch there.

README.md
```diff
@@ -18,3 +18,28 @@ These expectations are as follows:
 - The `physical_dimension` field for any given `EDF.Signal` is a value supported by `OndaEDF.STANDARD_UNITS`.
 
 Note that callers can additionally use the `custom_extractors` argument to `edf_to_onda_signals` to workaround some of these expectations; see the `import_edf` docstring for more details.
+
+## Fine-grained control over `Signal` processing: `plan` and `execute_plan`
```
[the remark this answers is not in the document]
kleinschmidt: I think I got everything in the docstrings, this was the last thing I'd missed.

src/import_edf.jl
palday: if we didn't, we should, because strictly speaking M1 is the left mastoid and A1 is the left earlobe (even though a lot of labs used them interchangeably)

src/OndaEDF.jl
```diff
@@ -1,26 +1,131 @@
...
+                kind::Union{Missing, AbstractString}
+                channel::Union{Missing, AbstractString}
+                sample_unit::Union{Missing, AbstractString}
+                sample_resolution_in_unit::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                sample_offset_in_unit::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                sample_type::Union{Missing, AbstractString}
+                sample_rate::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                # errors, use `nothing` to indicate no error
+                error::Union{Nothing, Exception}
+                )
```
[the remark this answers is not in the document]
kleinschmidt: these ones definitely are just copied straight from onda: https://github.com/beacon-biosignals/Onda.jl/blob/main/src/signals.jl#L62-L69

src/OndaEDF.jl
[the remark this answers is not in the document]
ararslan: `convert`s already happen automatically on construction of the row object:
```
julia> using Legolas

julia> struct Meme end

julia> Base.convert(::Type{Meme}, x::Int) = (println("gottem"); Meme())

julia> Bean = Legolas.@row("bean@1", meme::Meme)
Legolas.Row{Legolas.Schema{:bean, 1}, F} where F

julia> Bean((; meme=1))
gottem
Row(Schema("bean@1"), (meme = Meme(),))
```

src/import_edf.jl
ararslan: I'm confused. If someone wants to group things differently then aren't the previously computed indices already irrelevant?

src/import_edf.jl
ararslan: If we accept a signal header or a corresponding `NamedTuple` then I think it's reasonable for callers to do their own custom preprocessing and pass the result to this function rather than handling the calling for them. Now _that_'s inverting control.

src/import_edf.jl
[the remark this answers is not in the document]
ararslan: :ok_hand:

src/import_edf.jl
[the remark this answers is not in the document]
kleinschmidt: I'm gonna go ahead and make that change, error is now a string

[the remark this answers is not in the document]
ararslan: That's not too surprising. We could do something like
```julia
function stringify_exception(ex, bt=catch_backtrace())
    io = IOBuffer()
    showerror(io, ex)
    if !isempty(bt)
        st = stacktrace(bt)
        println(io, '\n', "Stacktrace:")
        join(io, ("  [$i] $frame" for (i, frame) in enumerate(st)), '\n')
    end
    return String(take!(io))
end
```
Which will give you the full stack trace in the same string as the error message.

src/OndaEDF.jl
```diff
@@ -1,26 +1,131 @@
...
 
 using Base.Iterators
-using Compat: @compat
-using Dates, UUIDs
-using Onda, EDF
+using Dates
+using EDF
+using Legolas
+using Onda
+using PrettyTables
```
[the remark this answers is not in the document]
ararslan: You can do your own minimal writing out of `NamedTuple`s, seems fairly straightforward.

src/import_edf.jl
kleinschmidt: they might be MOSTLY correct but need a little tweak.  you can then insert another index column that gets combined with the original index, so e.g. everywhere 1 but a few 2s to separate out a few channels into distinct samples from the other channels with the same `onda_signal_index`,

src/import_edf.jl
kleinschmidt: I'm pretty sure we did remove that, I'll update teh docstring

src/OndaEDF.jl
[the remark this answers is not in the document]
kleinschmidt: yeah but with this you get nice alignment of columns too which is a big plus.  sometimes you need to be able to see a few rows at once and see where they differ (in `merge_samples_info`, you need to see what the non-unique values of the key are)

src/import_edf.jl
kleinschmidt: that might be a good test case actually...

src/OndaEDF.jl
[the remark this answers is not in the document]
kleinschmidt: huh, TIL.  these are all based on what Onda.jl does.  I'll take em out anywy tho

src/import_edf.jl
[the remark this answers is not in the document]
kleinschmidt: that's basically what `showerror(io, e, bt)` does :)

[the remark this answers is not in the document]
kleinschmidt: (except the string part)

src/OndaEDF.jl
```diff
@@ -1,26 +1,129 @@
...
+                # Onda.SamplesInfo fields (channels -> channel), may be missing
+                kind::Union{Missing, AbstractString}
+                channel::Union{Missing, AbstractString}
+                sample_unit::Union{Missing, AbstractString}
+                sample_resolution_in_unit::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                sample_offset_in_unit::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                sample_type::Union{Missing, AbstractString}
+                sample_rate::Union{Missing, LPCM_SAMPLE_TYPE_UNION}
+                # errors, use `nothing` to indicate no error
+                error::Union{Nothing, String})
```
alexmchan, line 50: The documentation of these schemas might be useful to have in the `README.md` to make it easier to find?

alexmchan (approved):
LGTM

src/OndaEDF.jl
palday: The README covers their contents in way, but I would still add a note about their name (both `Plan` and `ondaedf.plan`) for breadcrumbs on where to look for the full details. :smile:

kleinschmidt: added an explicit call out for the schemas and to `write_plan`.

src/import_edf.jl
```diff
@@ -114,8 +203,18 @@ function edf_signal_encoding(edf_signal_header::EDF.SignalHeader,
             sample_type=sample_type)
 end
 
+# TODO: replace this with float type for mismatched
 function promote_encodings(encodings; pick_offset=(_ -> 0.0), pick_resolution=minimum)
-    sample_type = reduce(promote_type, (e.sample_type for e in encodings))
+    if any(any(ismissing, row) for row in encodings)
+        return (sample_type=missing,
```
palday, line 209:
```suggestion
        return (; sample_type=missing,
```

src/import_edf.jl
```diff
@@ -208,81 +267,262 @@ function groupby(f, list)
     return d
 end
 
+# unpack a single channel spec from labels:
+# "channel"
+canonical_channel_name(channel_name) = channel_name
+# "channel" => ["alt1", "alt2", ...]
+canonical_channel_name(channel_alternates::Pair) = first(channel_alternates)
```
palday, line 274: I think this is mostly clear from the example above, but it would be good to have in the docs/docstrings somewhere that the specification is `canonical_name => [alt1, alt2]`. In other words, just documenting this behavior in a way that's discoverable without looking at the source.

palday: got a bit further and saw that this is documented

test/diagnostics/import.jl
```diff
@@ -0,0 +1,279 @@
...
+        results = map(test_edf_to_samples_info) do r
+            edf = mock_edf(r)
+            return last(OndaEDF.edf_header_to_onda_samples_info(edf; custom_extractors=custom_extractors))
+        end
+        # print result of mapping `edf_to_onda_samples` over `test_edf_to_samples_info.in`
+        # this makes it easy to see effects of any changes made to OndaEDF by looking at
+        # `diff test_edf_to_samples_info.{in,out}`
+        #
+        # in practice this doesn't work well because the ordering of dicts is not defined
+        # so the header map entries print out in different orders.
```
palday, line 267: so this is just testing that the code actually runs without error and not content?

Re ordering of dicts: you could define the ordering behavior to be based on `sort(keys(dict))` for consistency

palday (approved):

test/diagnostics/import.jl
kleinschmidt: none of the code in this directory is actually run anymore, just preserved for posterity

kleinschmidt: (and to  make it a bit easier if we want to restore the tests later)

palday: I had completely lost the thread by this point. :smile:

kleinschmidt: also (now that I'm thining about it again) because I wrote a bit of glue code to convert the old diagnostics into a table and vice-versa that might be useful.

src/import_edf.jl
kleinschmidt: Yeah, that's fair.  The issue is that it's a bit clunky then to do the file-level operations (like assigning `edf_signal_index` and `onda_signal_index`), since that's all handled by the `plan_edf_to_onda_samples(::EDF.File, ...)` method which just forwards it's kwargs to the `plan_edf_to_onda_samples(header, ...)` method.

src/import_edf.jl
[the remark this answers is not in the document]
ararslan: TIL there's a method that takes a backtrace as an argument, nice

src/import_edf.jl
kleinschmidt: I'm actually coming around to the idea that there should be no option to change which column gets used for this, but that we should refactor and possibly export the function that actually does the grouping/index writing, which would make it more ergonomic to work with the single-channel `plan` interface.  theeeennnn maybe we can talk about getting rid of the preprocessor :)

kleinschmidt: I've now refactored teh grouping stuff into its own function `plan_edf_to_onda_signal_groups` which takes a plan table and a (tuple/vector of) columns to groupby.  this will check whether the `edf_signal_index` column already exists (which it might not if someone has run `plan_edf_to_onda_samples` one signal at a time); if it doesn't it gets filled in based on the order of the rows passed in.  This introduces the possibility of errors if the rows have been re-ordered, but there is already a check (with a nice error message) that the signal label matches the plan label when doing the actual conversion into samples so I'm comfortable that that won't cause big problems.

kleinschmidt:
I've made some substantive changes to this based on @ararslan feedback, which mostly involve refactoring how EDF Files are planned to separate the "grouping" (edf/onda index setting) into its own function `plan_edf_to_onda_signal_groups` which is now exported.  I chose to export it because it's necessary for supporting workflows where the signal header is modified before being passed to plan.  I've also added a few scenarios for this kind of workflows in the README, as well as tests for them.

The last decision point is whether to completely remove the `preprocess_labels` kwarg from the planning functions.  I'm in favor of that at this point since we're already making some pretty breaking changes, but it might be better to have a deprecation cycle before we do that.

README.md
```diff
@@ -17,4 +17,108 @@ These expectations are as follows:
...
+    header = signal.header
+    return Tables.rowmerge(header; 
+                           label=header.transducer_type, 
+                           transducer_type=header.label)
+end
+
+plans = map(plan_edf_to_onda_samples ∘ corrected_header, edf.signals)
+grouped_plans = OndaEDF.plan_edf_to_onda_signal_groups(plans)
+samples, plan_executed = edf_to_onda_samples(edf, grouped_plans)
+```
```
kleinschmidt, line 124: this scenario shows how a user can preprocess the signal headers before planning

README.md
```diff
@@ -17,4 +17,108 @@ These expectations are as follows:
...
+    else
+        return plan
+    end
+end
+
+new_plan = map(fix_emg, Tables.rows(plans))
+# re-compute the grouping of EDF signals into Onda signals:
+new_plan = OndaEDF.plan_edf_to_onda_signal_groups(new_plan)
+samples, plan_executed = edf_to_onda_samples(edf, new_plan)
+```
```
kleinschmidt, line 103: this scenario shows a more complicated post-processing where you try to merge emg_1 and emg_2 etc., which requires re-grouping

test/import.jl
```diff
@@ -1,325 +1,89 @@
...
-            @test samples_info["sao2"].sample_unit == "percent"
-            @test samples_info["ptaf"].channels == ["ptaf"]
-            @test samples_info["ptaf"].sample_unit == "volt"
-
+            validate_extracted_signals(s.info for s in returned_samples)
             @test all(Onda.duration(s) == Nanosecond(Second(200)) for s in returned_samples)
         end
     end
 
+    @testset "edf_to_onda_samples with manual override" begin
```
kleinschmidt, line 21: these tests cover row-wise plans and manual use of the grouping function

kleinschmidt:
Ugh I pushed 4ff575b to this branch but it's not showing up in the PR for some reason.  I'll hold off on merging until that's resolved (I'm assuming it's somehow related to the github weirdness today)

test/diagnostics/import.jl
ararslan: If these tests aren't being run at all then I would delete them. They're available in the git history if we ever want to restore them but in the meantime it's confusing to have a bunch of bitrotten tests lying around unused.

test/import.jl
```diff
@@ -1,325 +1,89 @@
...
+            signal_plans = [rowmerge(plan; grp=string(plan.kind, plan.sample_unit, plan.sample_rate))
+                            for plan in signal_plans]
+            grouped_plans = plan_edf_to_onda_samples_groups(signal_plans,
+                                                                    onda_signal_groupby=:grp)
+            returned_samples, plan = edf_to_onda_samples(edf, grouped_plans)
+            validate_extracted_signals(s.info for s in returned_samples)
+
+            # one channel per signal, group by label
+            grouped_plans = plan_edf_to_onda_samples_groups(signal_plans,
+                                                                    onda_signal_groupby=:label)
```
ararslan, line 47:
```suggestion
            grouped_plans = plan_edf_to_onda_samples_groups(signal_plans,
                                                            onda_signal_groupby=:grp)
            returned_samples, plan = edf_to_onda_samples(edf, grouped_plans)
            validate_extracted_signals(s.info for s in returned_samples)

            # one channel per signal, group by label
            grouped_plans = plan_edf_to_onda_samples_groups(signal_plans,
                                                            onda_signal_groupby=:label)
```

src/import_edf.jl
```diff
@@ -100,22 +190,29 @@ end
...
     sample_rate = edf_signal_header.samples_per_record / edf_seconds_per_record
-    sample_type = (dmax > typemax(Int16) || dmin < typemin(Int16)) ? Int32 : Int16
+    sample_type = (dmax > typemax(Int16) || dmin < typemin(Int16)) ? "int32" : "int16"
     return (sample_resolution_in_unit=Float64(sample_resolution_in_unit),
             sample_offset_in_unit=Float64(sample_offset_in_unit),
             sample_rate=Float64(sample_rate),
             sample_type=sample_type)
 end
 
+# TODO: replace this with float type for mismatched
```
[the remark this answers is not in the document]
kleinschmidt: for future us; didn't want to make this any gnarlier than it already was.

src/import_edf.jl
```diff
@@ -1,3 +1,36 @@
...
+    return :(NamedTuple{$names,$types}($body))
+end
+
+function _err_msg(e, msg="Error while converting EDF:")
+    bt = catch_backtrace()
+    msg = let io = IOBuffer()
+        println(io, msg)
+        showerror(io, e, bt)
+        String(take!(io))
+    end
```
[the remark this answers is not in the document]
kleinschmidt: `n i c e`

src/import_edf.jl
```diff
@@ -208,81 +265,314 @@ function groupby(f, list)
...
+to combine the EDF signals data into a single `Onda.Samples` per group.
+
+The `label` of the original `EDF.Signal`s are preserved in the `:edf_channels`
+field of the resulting `SamplesInfo`s for each `Samples` generated.
+
+Any errors that occur are shown as `String`s (with backtrace) and inserted into
+the `:error` column for the corresponding rows from the plan.
+
+Samples are returned in the order of `:onda_signal_index` (or otherwise the output
+of the `samples_groups` function).  Signals that could not be matched or
```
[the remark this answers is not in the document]
kleinschmidt: dead

src/import_edf.jl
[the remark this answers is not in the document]
kleinschmidt: it's also tracked in #36

src/import_edf.jl
```diff
@@ -208,81 +265,314 @@ function groupby(f, list)
...
+        rows = grouped_rows[key]
+        encoding = promote_encodings(rows)
+        return [rowmerge(row, encoding, (; onda_signal_index)) for row in rows]
+    end
+    return plan_rows
+end
+
+_get(x, property) = hasproperty(x, property) ? getproperty(x, property) : missing
+function grouper(vars=(:kind, :sample_unit, :sample_rate))
+    return x -> NamedTuple{vars}(_get.(Ref(x), vars))
```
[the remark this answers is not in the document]
kleinschmidt: :smiling_imp:

ararslan (approved):
Absolutely heroic.

PR #46, opened 2022-06-07
Use singular in full-service path

palday:
Legolas schema use singular by convention and Onda follows this convention. 

I'm unsure whether this should be considered a breaking change and hence what the necessary version bump is. In some sense it's a bug fix, but it definitely changes behavior in a non compatible way.

### DIFF
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -634,8 +634,8 @@ end
 
 Convert an EDF.File to `Onda.Samples` and `Onda.Annotation`s, store the samples
 in `\$path/samples/`, and write the Onda signals and annotations tables to
-`\$path/\$(signals_prefix).onda.signals.arrow` and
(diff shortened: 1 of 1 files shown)


jrevels:
> I'm unsure whether this should be considered a breaking change and hence what the necessary version bump is.

i'd consider it breaking (and thus entail a major bump) even though it's a bugfix since the likelihood of downstream users/code hardcoding these extensions is pretty high

given that it's a breaking change anyway I think I'd be a fan of just removing the magical extension handling bits altogether and changing the interface from e.g. `signal_prefix` to `signal_file_path`

better to just let the caller fully choose their own file paths/names IMO

kleinschmidt (commented):
I do think that for the "full service" option it's good to have defaults, but yeah let's make them file names instead of prefixes (if we're making a breaking change anyway)

Issue #50, opened 2022-07-18
schema definitions should be split into separate package


kleinschmidt:
In order for the schemas defined here to be used in any downstream package, they have to take OndaEDF as a depedendency, even if they just want to register that say a `"ondaedf.file-plan@1"` table is part of a dataset, or just load these tables to later analysis.  OndaEDF is pretty hefty dependency for these "dataset packages" which only track persistent identifiers of where tables with different schemas can be found.  We can fix this by splitting out the schema definitions in this package into an OndaEDFSchemas.jl package that's a "light" dependency.

ericphanson:
Yeah, I think if you want to be able to make a PR that updates two packages simulatenously/atomically, then a subpackage is good. I think schemas are a great example of that. SnoopCompile / SnoopCompileCore (https://github.com/timholy/SnoopCompile.jl) is another example of this, or Makie + it's backends.

kleinschmidt:
done with #51 and #53

PR #51, opened 2022-07-18
Create separate OndaEDFSchemas sub-package

kleinschmidt:
This partially addresses #50 by creating OndaEDFSchemas.jl inside this repo and adding tests, CI, and codecove for it.  We will need to wait to move OndaEDF.jl over to use this until it is registered though to avoid teh ArrowTypes situation.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,4 @@
 flag_management:
+  default_rules: # the rules that will be followed for any flag added, generally
+    carryforward: true
 comment: off


codecov.yml
```diff
@@ -1 +1,10 @@
+flag_management:
+  default_rules: # the rules that will be followed for any flag added, generally
+    carryforward: true
+    statuses:
+      - type: project
+        target: auto
+        threshold: 1%
+      - type: patch
+        target: 90%
 comment: off
```
palday, line 10: do we need this or can we just take advantage of org settings?

palday (approved):

codecov.yml
kleinschmidt: we need at least the `flag_management` thing, let me try without that...

kleinschmidt: null value not allowed there but I can ditch all teh statuses: bits

kleinschmidt:
Gonna merge this and worry about getting CI perfect later :)

PR #52, opened 2022-07-19
Add license and julia compat for OndaEDFSchemas.jl

kleinschmidt:
For registration

### DIFF
--- a/OndaEDFSchemas.jl/LICENSE.md
+++ b/OndaEDFSchemas.jl/LICENSE.md
@@ -1,6 +1,6 @@
 The OndaEDFSchemas.jl package is licensed under the MIT "Expat" License:
 
-> Copyright (c) 2020: Beacon Biosignals, Inc.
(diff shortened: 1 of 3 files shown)


OndaEDFSchemas.jl/LICENSE.md
```diff
@@ -0,0 +1,22 @@
+The OndaEDFSchemas.jl package is licensed under the MIT "Expat" License:
+
+> Copyright (c) 2020: Beacon Biosignals, Inc.
```
palday, line 3: is `2020` right? I suspect it should be `2020-2022`.

palday (approved):

OndaEDFSchemas.jl/LICENSE.md
kleinschmidt: hmmm, good q.  Technically this package was only "released" in 2022 so it should probably just be 2022

PR #53, opened 2022-07-19
Use registered OndaEDFSchemas

kleinschmidt:


palday (approved):

kleinschmidt:
Something is stil not quite right about the coverage...I'll attempt to post a fix that reverts some of the changes from #51

PR #55, opened 2022-07-19
use EDFAnnotation schema for import/export

### LINKED ISSUE #54: export assumes `value` field is present in the annotations
In the brave new world of Legolas schemas for annotations, the annotations table is only required to have `recording`, `id`, and `span` columns.  Previous iterations of Beacon's annotation format also had a `value` field, but that's no longer universally true.  The current export code fails on those tables (where the relevant "value" information is recorded in a field with a different name).  I think what we should do is

- Implement a schema for EDF annotations as generated during _import_
- Implement a schema for the minimum fields required for _export_, and validate the input annotations during export.

From a very quick perusal of the code, the same schema could be used for both (the import puts the the zero or more strings in `.annotations` field of the EDF annotation into the `value` field of one or more `Onda.Annotation`s), so we could define

```julia
Legolas.@row("edf.annotation@1" > "onda.annotation@1", value::String=string(value))
```

kleinschmidt:
This fixes #54 by enforcing this assumption with a schema (`edf.annotation`) which extends `onda.annotation` with one extra field `value::String`.  It also changes the return type of imported EDF annotation to these rows (although the _values_ are identical); while this is technically breaking, it seems pretty harmless.


.github/workflows/CI.yml
```diff
@@ -38,6 +38,7 @@ jobs:
           path: ~/.julia/artifacts
           key: ${{ runner.os }}-test-artifacts-${{ hashFiles('**/Project.toml') }}
           restore-keys: ${{ runner.os }}-test-artifacts
+      - run: julia --project -e 'using Pkg; Pkg.develop([PackageSpec(; path="OndaEDFSchemas.jl")])'
```
omus, line 41: Should only be needed until OndaEDFSchemas.jl is registered. My personal preference would be to have this work split into two PRs. Also, I do find using a committed Manifest.toml is a cleaner way to do that but mainly that's only useful while developing the PR

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -3,7 +3,18 @@ module OndaEDFSchemas
...
+
+"""
+    EDFAnnotation = @row("edf.annotation@1" > "onda.annotation@1",
+                         value::String = string(value))
+
+A type-alias for a Legolas row that represents one annotation extracted from an
+EDF file.  The only extra field is `value`, which holds the string stored in the
+elements of `annotations` in the EDF annotation.
+"""
+const EDFAnnotation = @row("edf.annotation@1" > "onda.annotation@1",
```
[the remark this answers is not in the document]
kleinschmidt: I think my intention here was that this is an annotation that is derived from an EDF Annotation signal.  The fact that it's imported by OndaEDF seems a bit beside the point (unlike the plan which is intrinsically an OndaEDF _thing_)

.github/workflows/CI.yml
kleinschmidt: I _think_ it is registered at this point.  Without this the registered version is picked up, which means that the OndaEDF CI will not pick up the OndaEDFSchemas changes in a PR.

kleinschmidt:
we're going to do this as part of the Legolas 0.5 upgrade #60

PR #57, opened 2022-09-16
Ignore `-org` as an EDF channel name suffix

ararslan:
I've encountered this quite a bit lately. It does not appear to be particularly meaningful in the same way that e.g. the `-ref` suffix isn't. No one is certain what exactly it means but the best guess is an abbreviation of "original" for "original reference."


kleinschmidt:
I think in this case, since it's not part of the spec (unlike `-Ref`), it would be better to manually preprocess the headers to remove it.

ararslan:
Oh right, the EDF spec

PR #58, opened 2022-09-20
Fix typo/extra space with the `etco2` channel name

kimlaberinto:
Otherwise the channel name "etco2 " will hit the invalid channel name error at:
https://github.com/beacon-biosignals/Onda.jl/blob/6952dd2b59fd16c02f0a80b96ca7c2f627b04ca0/src/signals.jl#L136

Credit to @mich11 for finding this problem in the source code!

### DIFF
--- a/src/standards.jl
+++ b/src/standards.jl
@@ -99,4 +99,4 @@ const STANDARD_LABELS = Dict(# This EEG channel name list is a combined 10/20 an
                              ["tidal_volume"] => ["tidal_volume"=> ["tvol", "tidal"]],
                              ["spo2"] => ["spo2"],
                              ["sao2"] => ["sao2", "osat"],
-                             ["etco2"] => ["etco2 "=> ["capno"]])
+                             ["etco2"] => ["etco2"=> ["capno"]])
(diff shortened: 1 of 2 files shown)


src/standards.jl
```diff
@@ -99,4 +99,4 @@ const STANDARD_LABELS = Dict(# This EEG channel name list is a combined 10/20 an
                              ["tidal_volume"] => ["tidal_volume"=> ["tvol", "tidal"]],
                              ["spo2"] => ["spo2"],
                              ["sao2"] => ["sao2", "osat"],
-                             ["etco2"] => ["etco2 "=> ["capno"]])
+                             ["etco2"] => ["etco2"=> ["capno"]])
```
kleinschmidt, line 102: make it YASG compliant while we're at it 😁 

```suggestion
                             ["etco2"] => ["etco2" => ["capno"]])
```

kleinschmidt (approved):
Thanks for the fix!  Sounds like we may need better test coverage here so if you feel like adding a test case that would be great.  And bump the patch version as well.

kimlaberinto:
Thank you for the review! Not able to get test case right now, but this PR should now be good to merge

Issue #60, opened 2022-11-30
upgrade to Legolas 0.5


kleinschmidt:
We define schemas for the EDF import plans (in OndaEDFSchemas.jl sub-package), which needs to be upgraded to Legolas 0.5.  Nothing in those schemas requires any breaking changes so we should be able to preserve backward compatibility with existing plan tables.

We also need to upgrade the emitted signal/annotation tables to be Onda 0.15/Legolas 0.5 compatible as well.

In particular, the more strict stance of Legolas 0.5 on schema extensions means that we _will_ need something like #55 to emit annotation schemas with an extra `value` field.

Also, the Signal/SamplesInfo schemas in Onda have changed in ways that impact the way we import EDFs:
- `kind` replaced with `sensor_type` and `sensor_label`

PR #61, opened 2022-12-06
Upgrade OndaEDFSchemas to Legolas 0.5, Onda 0.15

christopher-dG:
this is half of #60

### DIFF
--- a/OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
+++ b/OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
@@ -136,7 +136,7 @@ const FILE_PLAN_DOC_TEMPLATE = """
         onda_signal_index::Int
     end
 
-Type alias for a Legolas row for one EDF signal-to-Onda channel conversion,
+A Legolas-generated record type representing one EDF signal-to-Onda channel conversion,
 which includes the columns of a [`PlanV{{ VERSION }}`](@ref) and additional file-level context:
 - `edf_signal_index` gives the index of the `signals` in the source `EDF.File`
   corresponding to this row
--- a/OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
+++ b/OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
@@ -1,7 +1,9 @@
 module OndaEDFSchemas
 
 using Legolas: Legolas, @schema, @version, lift
-using Onda: LPCM_SAMPLE_TYPE_UNION, onda_sample_type_from_julia_type, convert_number_to_lpcm_sample_type
+using Onda: LPCM_SAMPLE_TYPE_UNION, onda_sample_type_from_julia_type,
+            convert_number_to_lpcm_sample_type, _validate_signal_channel,
+            _validate_signal_sensor_label, _validate_signal_sensor_type
 
 export PlanV1, PlanV2, FilePlanV1, FilePlanV2
 
@@ -32,6 +34,8 @@ export PlanV1, PlanV2, FilePlanV1, FilePlanV2
     error::Union{Nothing,String} = coalesce(error, nothing)
 end
 
+Legolas.accepted_field_type(::PlanV1SchemaVersion, ::Type{String}) = AbstractString
+
 @version PlanV2 begin
     # EDF.SignalHeader fields
     label::String
@@ -46,9 +50,9 @@ end
     # EDF.FileHeader field
     seconds_per_record::Float64
     # Onda.SamplesInfoV2 fields (channels -> channel), may be missing
-    sensor_type::Union{Missing,AbstractString} = lift(String, sensor_type)
-    sensor_label::Union{Missing,AbstractString} = lift(String, sensor_type)
-    channel::Union{Missing,AbstractString} = lift(String, channel)
+    sensor_type::Union{Missing,AbstractString} = lift(_validate_signal_sensor_type, sensor_type)
+    sensor_label::Union{Missing,AbstractString} = lift(_validate_signal_sensor_label, sensor_type)
+    channel::Union{Missing,AbstractString} = lift(_validate_signal_channel, channel)
     sample_unit::Union{Missing,AbstractString} = lift(String, sample_unit)
     sample_resolution_in_unit::Union{Missing,Float64}
     sample_offset_in_unit::Union{Missing,Float64}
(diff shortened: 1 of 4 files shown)


OndaEDFSchemas.jl/test/runtests.jl
```diff
@@ -7,61 +7,73 @@ using StableRNGs
...
-    tbl = Arrow.Table(Arrow.tobuffer(plans))
-    @test isequal(Tables.columntable(tbl), Tables.columntable(plans))
-end
+for v in [1, 2]
+    @testset "ondaedf.plan@$v" begin
+        rng = StableRNG(10)
+        plans = mock_plan(30; v, rng)
+        schema = Tables.schema(plans)
+        @test nothing === Legolas.validate(schema, Legolas.SchemaVersion("ondaedf.plan", v))
+        tbl = Arrow.Table(Arrow.tobuffer(plans; maxdepth=9))
```
kleinschmidt, line 65: it's a bit odd that this is now necessary (even for the V1 schema)...can you maybe look through the CI runs to find the last time this ran successfully.  is there some reason why Legolas 0.5 would have changed something here?

OndaEDFSchemas.jl/test/runtests.jl
```diff
@@ -7,61 +7,73 @@ using StableRNGs
...
 end
 
-@testset "ondaedf.plan@1" begin
-    rng = StableRNG(10)
-    plans = mock_plan(30; rng)
-    @test nothing === Legolas.validate(plans, Legolas.Schema("ondaedf.plan@1"))
-    tbl = Arrow.Table(Arrow.tobuffer(plans))
-    @test isequal(Tables.columntable(tbl), Tables.columntable(plans))
-end
+for v in [1, 2]
```
kleinschmidt, line 59: you _could_ do this as a testset if you wanted...

```suggestion
@testset "Schema version $v" for v in (1, 2)
```

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -1,94 +1,135 @@
...
+    # EDF.FileHeader field
+    seconds_per_record::Float64
+    # Onda.SamplesInfo fields (channels -> channel), may be missing
+    sensor_type::Union{Missing,AbstractString} = lift(String, sensor_type)
+    channel::Union{Missing,AbstractString} = lift(String, channel)
+    sample_unit::Union{Missing,AbstractString} = lift(String, sample_unit)
+    sample_resolution_in_unit::Union{Missing,LPCM_SAMPLE_TYPE_UNION} = lift(convert_number_to_lpcm_sample_type, sample_resolution_in_unit)
+    sample_offset_in_unit::Union{Missing,LPCM_SAMPLE_TYPE_UNION} = lift(convert_number_to_lpcm_sample_type, sample_offset_in_unit)
+    sample_type::Union{Missing,AbstractString} = lift(onda_sample_type_from_julia_type, sample_type)
+    sample_rate::Union{Missing,LPCM_SAMPLE_TYPE_UNION} = lift(convert_number_to_lpcm_sample_type, sample_rate)
```
kleinschmidt, line 55: I think these fields have different types now in SignalV2: 

https://github.com/beacon-biosignals/Onda.jl/blob/1c69482315700342c706c1c4d4800231c0bc93e9/src/signals.jl#L80-L83

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -1,94 +1,135 @@
...
+        kind::Union{Missing,AbstractString}
+        channel::Union{Missing,AbstractString}
+        sample_unit::Union{Missing,AbstractString}
+        sample_resolution_in_unit::Union{Missing,LPCM_SAMPLE_TYPE_UNION}
+        sample_offset_in_unit::Union{Missing,LPCM_SAMPLE_TYPE_UNION}
+        sample_type::Union{Missing,AbstractString}
+        sample_rate::Union{Missing,LPCM_SAMPLE_TYPE_UNION}
+        # errors, use `nothing` to indicate no error
+        error::Union{Nothing,String}
+    end
```
kleinschmidt, line 84: becuse the field names and types are different enough, might need to create two separate docstrings (maybe with some of hte common stuff templated out?)

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -1,94 +1,135 @@
...
+    physical_minimum::Float32
+    physical_maximum::Float32
+    digital_minimum::Float32
+    digital_maximum::Float32
+    prefilter::String
+    samples_per_record::Int16
+    # EDF.FileHeader field
+    seconds_per_record::Float64
+    # Onda.SamplesInfo fields (channels -> channel), may be missing
+    sensor_type::Union{Missing,AbstractString} = lift(String, sensor_type)
```
kleinschmidt, line 49: also need `sensor_label`:

https://github.com/beacon-biosignals/Onda.jl/blob/1c69482315700342c706c1c4d4800231c0bc93e9/src/signals.jl#L121-L130

kleinschmidt (changes requested):
good start!  the serialization depth thing is concerning to me, and I dont' really know what might be causing that now that wasn't before.  either something in legolas 0.5 has changed the serialization behavior, or arrow/arrowtypes has had an upgrade that caused some breakage there, oorrrrr something else more mysterious.

otherwise, some of the V2 fields need tweaking I think.  will likely need ot `using` more stuff from Onda for validating these fields (see the linked Onda source).

last thing is that I think we need to #55 first/include it in here, since `Onda.AnnotationV1` may now just drop the extra `value` field we create during import.

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
christopher-dG: oops, yes definitely

OndaEDFSchemas.jl/test/runtests.jl
christopher-dG: @kleinschmidt not entirely sure what you mean by checking CI logs. The last run was [3 months ago](https://github.com/beacon-biosignals/OndaEDF.jl/actions/runs/3093327014/jobs/5005573529) on Onda 0.14 / Legolas 0.4 so I don't know that that tells us anything

kleinschmidt: the only possibly relevant difference I can see is that that run was on Arrow 2.3.1 and we're getting 2.4 now

kleinschmidt (approved):

PR #62, opened 2022-12-10
upgrade OndaEDF import/export to legolas 0.5/onda 0.15

kleinschmidt:
This upgrades the import/export functionality to Legolas 0.5 and Onda 0.15.  Builds on #61 (and fixes a small issue there with arrow introducing `Missing` into `Union{Nothing,T}` columns).

I've really done the bare minimum here so far:
- use updated schemas from #61 (including `EDFAnnotationV1` instead of `Onda.Annotation`, due to Legolas 0.5 schemas dropping extra columns)
- update SamplesInfoV2, using `sensor_type` in place of `kind`.

Still TODO:
- [x] ~~generate uniform `sensor_type` and unique `sensor_label` when EDF signals with teh same `sensor_type` cannot be combined into a single  (this may need tests actually)~~ see below https://github.com/beacon-biosignals/OndaEDF.jl/pull/62#issuecomment-1349346867
- [x] update examples in README
- [x] check that docs are up-to-date (need to also `dev` OndaEDFSchemas as well)


kleinschmidt:
Okay, so apparently the README is a lie (who wrote this crap) and we don't actually generate unique `kind` entries on main.  I've added a test to confirm this.  How to handle the `sensor_label` metadata is a bit tricky actually, since it doesn't actually get included in the `SamplesInfoV2` but only at the `SignalV2` stage (e.g., after `Onda.store(::Samples, metadata...)`).  Given the design decision to _not_ provide a `store` hook, I think it's basically up to the caller to extract the `sensor_label` from teh generated plan table and add it ot the call to `Onda.store` themselves.  I'm not very happy with that (it makes the `edf_to_onda_samples` abstraction leaky and requires that the user also do further aggregation of the plan in to signal-level), and makes me think it was a mistake to include the `sensor_label` in the plan tables in the first place if we're not doing anything meaningful with it.  But "doing somehting meaningful with `sensor_label`" is I think a bit out of scope for this PR and shoudl be left to a followup...

omus (commented):
Partial review (sorry). I'll try to finish reviewing later today

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -36,6 +36,9 @@ export PlanV1, PlanV2, FilePlanV1, FilePlanV2, EDFAnnotationV1
 end
 
 Legolas.accepted_field_type(::PlanV1SchemaVersion, ::Type{String}) = AbstractString
+# we need this because Arrow write can introduce a Missing for the error column
+# (I think because of how missing/nothing sentinels are handled?)
+Legolas.accepted_field_type(::PlanV1SchemaVersion, ::Type{Union{Nothing,String}}) = Union{Nothing,Missing,AbstractString}
```
omus, line 41: Let's try to reduce the code duplication:
```suggestion
const OndaEDFSchemaVersions = Union{PlanV1SchemaVersion,PlanV2SchemaVersion,FilePlanV1SchemaVersion,FilePlanV1SchemaVersion}
Legolas.accepted_field_type(::OndaEDFSchemaVersions, ::Type{String}) = AbstractString
# we need this because Arrow write can introduce a Missing for the error column
# (I think because of how missing/nothing sentinels are handled?)
Legolas.accepted_field_type(::OndaEDFSchemaVersions, ::Type{Union{Nothing,String}}) = Union{Nothing,Missing,AbstractString}
```

README.md
```diff
@@ -59,7 +59,7 @@ edf = EDF.File(my_edf_file_path)
 plans = plan_edf_to_onda_samples(edf; label=my_labels)
 
 function fix_millivolts(plan)
-    if plan.sample_unit == "millivolt" && plan.kind == "eeg"
+    if plan.sample_unit == "millivolt" && plan.sensor_type == "eeg"
```
omus, line 62: Were these untested? How did we do the initial upgrade to Legolas 0.5 without this failing?

README.md
```diff
@@ -75,23 +75,22 @@ samples, plan_executed = edf_to_onda_samples(edf, new_plan)
...
+    if plan.sensor_type == "emg"
         if plan.sample_unit == "millivolt"
             sample_resolution_in_unit = plan.sample_resolution_in_unit * 1000
             sample_offset_in_unit = plan.sample_offset_in_unit * 1000
             plan = Tables.rowmerge(plan; sample_unit="microvolt",
                                    sample_resolution_in_unit,
                                    sample_offset_in_unit)
         end
-        return Tables.rowmerge(plan; kind="emg")
+        return Tables.rowmerge(plan; sensor_label="emg")
```
omus, line 93: Is the label rename still required?

test/import.jl
```diff
@@ -193,6 +193,31 @@ using Legolas: validate, Schema, read
         end
     end
 
+    @testset "duplicate sensor_type" begin
+        rng = MersenneTwister(1234)
```
omus, line 197: Better to use StableRNG.jl so this works consistently on different Julia versions

README.md
kleinschmidt: nope, good catch.

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -126,11 +134,15 @@ end
     onda_signal_index::Int
 end
 
+Legolas.accepted_field_type(::FilePlanV1SchemaVersion, ::Type{Union{Nothing,String}}) = Union{Nothing,Missing,AbstractString}
+
```
kleinschmidt, line 138:
```suggestion
```

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -126,11 +134,15 @@ end
...
 @version FilePlanV2 > PlanV2 begin
     edf_signal_index::Int
     onda_signal_index::Int
 end
 
+Legolas.accepted_field_type(::FilePlanV2SchemaVersion, ::Type{Union{Nothing,String}}) = Union{Nothing,Missing,AbstractString}
+
```
kleinschmidt, line 145:
```suggestion
```

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -64,6 +68,10 @@ Legolas.accepted_field_type(::PlanV1SchemaVersion, ::Type{String}) = AbstractStr
 end
 
 Legolas.accepted_field_type(::PlanV2SchemaVersion, ::Type{String}) = AbstractString
+# we need this because Arrow write can introduce a Missing for the error column
+# (I think because of how missing/nothing sentinels are handled?)
+Legolas.accepted_field_type(::PlanV2SchemaVersion, ::Type{Union{Nothing,String}}) = Union{Nothing,Missing,AbstractString}
```
kleinschmidt, line 73:
```suggestion
```

README.md
kleinschmidt: First, these are in OndaEDF, not OndaEDFSchemas (which is where the initial upgrade was confined to).

Second, these are just examples in the README, so no, they're not tested.

README.md
kleinschmidt, line 93:
```suggestion
        return plan
```

kleinschmidt:
Quick update here: I've opted to go with "integration mode" (i.e., `] dev ./OndaEDFSchemas.jl`) for CI since that's the only reasonable option in my view:
- when OndaEDFSchemas needs to change in order to support parallel changes in OndaEDF (like in this PR), CI _without_ dev-ing will always fail since updated compat bounds are incompatible with the released version
- when OndaEDFSchemas has not changed, using the dev-ed version is usually equivalent to using the latest released version (and if it's not we've been naughty)

so using the released version of OndaEDFSchemas in CI isn't ever adding any useful information, unless I've missed something!

OndaEDFSchemas.jl/src/OndaEDFSchemas.jl
```diff
@@ -170,7 +175,7 @@ end
 A Legolas-generated record type that represents a single annotation imported
 from an EDF Annotation signal.  The `value` field contains the annotation value
 as a string.
-
 """
+EDFAnnotationV1
```
omus, line 179: With the newer versions of Legolas you can attach the docstring directly on the `@version` macro usage

README.md
omus: READMEs

Project.toml
```diff
@@ -32,8 +32,9 @@ julia = "1.6"
 [extras]
 FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
omus, line 35: Include a compat version

.github/workflows/Documentation.yml
```diff
@@ -11,8 +11,15 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/julia-buildpkg@latest
+      - uses: actions/checkout@v3
+      - name: "Monorepo setup"
+        # also dev OndaEDFSchemas.jl for pull request builds
+        if: ${{ github.event_name == 'pull_request' }}
```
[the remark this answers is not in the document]
kleinschmidt: ah yeah good point.  alternatively we could use the local copy for both push and pull_request, and only disable it for tags.  that way we can still have dev docs (although given our release cadence I'm not sure how useful that is...)

.github/workflows/Documentation.yml
```diff
@@ -2,17 +2,22 @@ name: Documentation
...
   Documenter:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/julia-buildpkg@latest
+      - uses: actions/checkout@v3
+      - name: "Monorepo setup"
+        # also dev OndaEDFSchemas.jl for pull request builds
+        if: ${{ github.event_name == 'pull_request' }}
```
omus, line 15: Documentation built in PRs and on `main` will use the version of OndaEDFSchemas.jl used in this commit while the documentation built for a specific version will use only a released version of OndaEDFSchemas.jl:
```suggestion
        # Only use a released version of OndaEDFSchemas.jl when creating documentation for a tag
        if: ${{ github.event_name != 'push' || !startsWith(github.ref, 'refs/tags') }}
```

.github/workflows/Documentation.yml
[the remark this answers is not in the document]
omus: Suggestion: https://github.com/beacon-biosignals/OndaEDF.jl/pull/62/files#r1087245685

.github/workflows/Documentation.yml
```diff
@@ -2,17 +2,22 @@ name: Documentation
...
+        shell: julia --color=yes --project=docs {0}
+        run: |
+          using Pkg
+          Pkg.develop([PackageSpec(; path="./OndaEDFSchemas.jl"),
+                       PackageSpec(; path=".")])
```
omus, line 20: I think you want to separate these two package dev calls now:
```suggestion
          Pkg.develop(PackageSpec(; path="./OndaEDFSchemas.jl"))
      - name: "Use latest OndaEDF.jl"
        shell: julia --color=yes --project=docs {0}
        run: |
          using Pkg
          Pkg.develop(PackageSpec(; path="."))
```

.github/workflows/Documentation.yml
```diff
@@ -2,17 +2,22 @@ name: Documentation
...
   pull_request:
 jobs:
   Documenter:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/julia-buildpkg@latest
+      - uses: actions/checkout@v3
+      - name: "Monorepo setup"
```
omus, line 13:
```suggestion
      - name: "Update dependencies"
```

omus (approved):
Good to merge once the CI changes are addressed

kleinschmidt:
> I think you want to separate these two package dev calls now. Otherwise we end up building the last released version of OndaEDF documentation:

I think we do need to `dev` both at the same time if we `dev OndaEDF` because of compatibility constraints: https://github.com/beacon-biosignals/OndaEDF.jl/actions/runs/4010515243/jobs/6887105869#step:3:8

Worse case is we just double-dev OndaEDF.jl which doesn't seem so bad to me.

kleinschmidt:
Actually, removing everything except Documenter from the docs/Project.toml did the trick.

PR #68, opened 2023-03-18
Add LUE and RAE to standard EOG channels

ararslan:
These are common PSG locations for EOG electrodes: left under canthus (LUE) and right above canthus (RAE).


ararslan:
Nope, at least not that I've ever seen. Btw did you mean to approve rather than comment?

Issue #70, opened 2023-04-18
Unable to match label with `EEG_` (or `anything_`)


josephsdavid:
I have a header that looks a bit like this:

```julia
EDF.SignalHeader("EEG_C3-A2", "", "uV", -313.0f0, 313.0f0, -32768.0f0, 32767.0f0, "", 200)
```

specifically, the label has an `EEG_` prefix (same issue arises with `ECG_` etc.), and it cannot be matched unless I manually get rid of the `EEG_`.

josephsdavid:
interestingly, if instead i replace `_` with ` `, almost everything works as expected!

kleinschmidt:
Root cause of this is that the regex we use to separate the signal and label is

https://github.com/beacon-biosignals/OndaEDF.jl/blob/1ec9351236c185c625fa300db9499dff0f0c7d8b/src/import_edf.jl#L149


Specifically, `[\s,\]]*\s+` which matches zero or more whitespace, `,`, or `]`, followed by one or more whitespace.  I wonder if we should use something like `[\s,\]]*[\s-_]+` instead so that we'd match `EEG_` or `EEG-` even though those are not STRICTLY allowed by the spec.

kleinschmidt:
AFAICT the only thing we're using the extracted `signal` match group for is to throw it away if it matches one of the values in the set of labels being considered...

kleinschmidt:
The spec isn't _super_ clear on this.  The EDF+ spec talks about a "standard 'label' text" but it's not obvious that labels are REQUIRED to have this structure (some of the samples they provide do not)

> 2.1. The standard 'label' structure
> The header field 'label' offers 16 ASCII characters. The standard structure consists of three components, from left to right:
> 
> - Type of signal (for example EEG).
> - A space.
> - Specification of the sensor (for example Fpz-Cz).
>
> As in all fields, this standard text must be left justified in the 16-character field and then filled out with spaces. In this example, the standard label reads 'EEG Fpz-Cz      '. Further possibilities are listed in the [signals table](https://www.edfplus.info/specs/edftexts.html#signals) below.

that being said, it's pretty clear that the EDF+ "standard" has an opinion that the signal/sensor spec separator shoudl be a space, but in practice we don't always see that (e.g. the example above from an open sleep dataset...)

PR #71, opened 2023-04-18
allow `_` to separate signal and sensor spec in labels

kleinschmidt:
This is one possible fix to #70 .  I'm honestly not sure if this is a good idea or not; on the one hand, it does allow us to handle things like `"EEG_C3-A2"`, but on the other hand, the EDF+ spec is fairly clear that the signal/sensor separator should be a single space.  HOWEVER, we allow all kinds of other crap in there besides just a single space so I'm not _too_ bothered by deviating from that slightly.

The alternative would be to require users pre-process the `label` field into something spec-compliant enough to be matched with `OndaEDF.match_edf_label`.

### DIFF
--- a/src/import_edf.jl
+++ b/src/import_edf.jl
@@ -146,7 +146,7 @@ function match_edf_label(label, signal_names, channel_name, canonical_names)
     # - if the signal name itself contains whitespace or one of `",[]"`, it
     #   will not match.  the fix for this is to preprocess signal headers before
     #   `plan_edf_to_onda_samples` to normalize known instances (after reviewing the plan)
-    m = match(r"[\s\[,\]]*(?<signal>.+?)[\s,\]]*\s+(?<spec>.+)"i, label)
+    m = match(r"[\s\[,\]]*(?<signal>.+?)[\s,\]]*[\s_]+(?<spec>.+)"i, label)
     if !isnothing(m) && m[:signal] in signal_names
         label = m[:spec]
     end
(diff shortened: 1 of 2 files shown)


ararslan (approved):

kleinschmidt:
WTH I _did not_ touch anything related to export??

ararslan:
Looks like a round-tripping issue perhaps

kleinschmidt:
ah yeah, it's because the generated names for some of the signals have `_` in them which messes up the new regex (since the signal type capture group needs to use lazy matching):

```
julia> [e.header.label for e in exported_edf.signals if !(e isa EDF.AnnotationsSignal)]
25-element Vector{String}:
 "EEG F3-M2"
 "EEG F4-M1"
 "EEG C3-M2"
 "EEG O1-M2"
 "EEG C4-M1"
 "EEG O2-A1"
 "EEG FPZ-Ref"
 "EOG LEFT"
 "EOG RIGHT"
 "ECG AVR"
 "ECG AVL"
 "EMG LEFT_ANTERIOR_TIBIALIS"
 "EMG RIGHT_ANTERIOR_TIBIALIS"
 "EMG INTERCOSTAL"
 "HEART_RATE HEART_RATE"
 "TIDAL_VOLUME TIDAL_VOLUME"
 "RESPIRATORY_EFFORT CHEST"
 "RESPIRATORY_EFFORT ABDOMEN"
 "SNORE SNORE"
 "POSITIVE_AIRWAY_PRESSURE IPAP"
 "POSITIVE_AIRWAY_PRESSURE EPAP"
 "PAP_DEVICE_LEAK PAP_DEVICE_LEAK"
 "PAP_DEVICE_CFLOW PAP_DEVICE_CFLOW"
 "SAO2 SAO2"
 "PTAF PTAF"
```

PR #76, opened 2023-06-26
fix sample encoding for export

### LINKED ISSUE #75: cannot export Onda Samples with `sample_type = UInt16`
MWE:

```julia
julia> using Onda, OndaEDF

julia> info = SamplesInfoV2(; sensor_type="x", channels=["x"], sample_unit="microvolt", sample_resolution_in_unit=2.0, sample_offset_in_unit=1.0, sample_type="uint16", sample_rate=100)
SamplesInfoV2: (sensor_type = "x", channels = ["x"], sample_unit = "microvolt", sample_resolution_in_unit = 2.0, sample_offset_in_unit = 1.0, sample_type = "uint16", sample_rate = 100.0)

julia> data = rand(UInt16, 1, 100);

julia> samples = Samples(data, info, true)
Samples (00:00:01.000000000):
  info.sensor_type: "x"
  info.channels: ["x"]
  info.sample_unit: "microvolt"
  info.sample_resolution_in_unit: 2.0
  info.sample_offset_in_unit: 1.0
  sample_type(info): UInt16
  info.sample_rate: 100.0 Hz
  encoded: true
  data:
1×100 Matrix{UInt16}:
 0xa063  0xeac6  0x5196  0x4e78  0x7126  0x2f30  0xe677  0x5f01  0xd096  0x9273  …  0xb6fb  0x4672  0x9963  0xef0e  0xbeec  0x621b  0xe988  0x0416  0x436d

julia> onda_to_edf([samples])
ERROR: MethodError: Cannot `convert` an object of type 
  EDF.Signal{UInt16} to an object of type 
  Union{EDF.AnnotationsSignal, EDF.Signal{Int16}}

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:64

Stacktrace:
 [1] push!(a::Vector{Union{EDF.AnnotationsSignal, EDF.Signal{Int16}}}, item::EDF.Signal{UInt16})
   @ Base ./array.jl:1060
 [2] onda_samples_to_edf_signals(onda_samples::Vector{Samples{Matrix{UInt16}}}, seconds_per_record::Float64)
   @ OndaEDF ~/work/OndaEDF.jl/src/export_edf.jl:124
 [3] onda_to_edf(samples::Vector{Samples{Matrix{UInt16}}}, annotations::Vector{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ OndaEDF ~/work/OndaEDF.jl/src/export_edf.jl:156
 [4] onda_to_edf
   @ ~/work/OndaEDF.jl/src/export_edf.jl:154 [inlined]
 [5] onda_to_edf(samples::Vector{Samples{Matrix{UInt16}}})
   @ OndaEDF ~/work/OndaEDF.jl/src/export_edf.jl:154
 [6] top-level scope
   @ REPL[49]:1
```

From digging around a bit in export, I think I have a hunch about what happened here, which is that when EDF added support for BDF (which use Int24 storage), the signal data field of hte `EDF.Signal` struct was made parametric on the eltype, and the implicit `convert(Vector{Int16}, data)` behavior in the old constructor we were relying on no longer happens.

kleinschmidt:
fixes #75 and cleans up a lot of crufty old code that's gotten out of sync with both Onda and EDF.

It adds a new internal function `reencode_samples` that takes `Onda.Samples` and returns new samples that are encoded in an EDF-friendly way (either directly as `Int16`/the requested type, or as something that can be converted losslessly to that type).  It tries very hard to _not_ change the encoding if it's possible, based on the typemin/max of the input `sample_type`, adn failing that, the acutal min/max of the encoded values (for `<:Integer` sample types).  If it's not possible to losslessly represent the encoded values as `Int16`, it picks a new quantization setting based on the min/max of the _decoded_ values (fitting those to typemin/max of `Int16` or whatever teh desired type is) and returning the re-encoded samples.

It also adds much more thorough tests of the export behavior, testing every type that Onda declares is allowed in `Onda.LPCM_SAMPLE_TYPE_UNION`.

While adding tests, I found that it's not possible for EDF.jl to decode things where the physical min/max are anywhere near the Float32 min/max finite values, see https://github.com/beacon-biosignals/EDF.jl/issues/67  As a result, some of the re-encoding tests do not hit the true typemin/max of the Onda sample type, but I think that's okay.  We're still in much better shape here than we were.

I'm putting this on hold for now but want to get a draft PR up

---
  - https://app.asana.com/0/0/1204892258999694


src/export_edf.jl
```diff
@@ -96,18 +96,87 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
...
 
+"""
+    reencode_samples(samples::Samples, sample_type::Type{<:Integer}=Int16)
+
+Re-compute encoding parameters for `samples` so that they can be encoded as
+`sample_type`.  The default `sample_type` is `Int16` which is the target for EDF
+format.
+
+This uses the actual signal extrema, choosing a resolution/offset that maps them
+to `typemin(sample_type), typemax(sample_type`.
```
palday, line 107:
```suggestion
to `typemin(sample_type), typemax(sample_type)`.
```

src/export_edf.jl
```diff
@@ -96,18 +96,90 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
                           is_contiguous, edf_record_metadata(samples)...)
 end
 
+"""
+    reencode_samples(samples::Samples, sample_type::Type{<:Integer}=Int16)
+
+Encode `samples` so that they can be encoded as `sample_type`.  The default
```
palday, line 102:
```suggestion
Encode `samples` so that they can be stored as `sample_type`.  The default
```

the double `encode` feels weird

src/export_edf.jl
```diff
@@ -96,18 +96,90 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
...
+        if !samples.encoded
+            smin, smax = Onda.encode_sample.(Onda.sample_type(samples.info),
+                                             samples.info.sample_resolution_in_unit,
+                                             samples.info.sample_offset_in_unit,
+                                             (smin, smax))
+        end
+        if smin >= typemin(sample_type) && smax <= typemax(sample_type)
+            # XXX: we're being a bit clever here in order to not allocate a
+            # whole new sample array, plugging in the new sample_type, re-using
+            # the old encodoed samples data, and skipping validation.  this is
```
palday, line 141:
```suggestion
            # the old encoded samples data, and skipping validation.  this is
```

src/export_edf.jl
```diff
@@ -96,18 +96,90 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
...
+            # okay in _this specific context_ since we know we're actually
+            # converting everything to Int16 in the actual export.
+            samples = encode(samples)
+            new_info = SamplesInfoV2(Tables.rowmerge(samples.info; sample_type))
+            return Samples(samples.data, new_info, true; validate=false)
```
palday, line 146: mildly cursed, but makes sense

src/export_edf.jl
```diff
@@ -96,18 +96,90 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
...
+                                             samples.info.sample_resolution_in_unit,
+                                             samples.info.sample_offset_in_unit,
+                                             (smin, smax))
+        end
+        if smin >= typemin(sample_type) && smax <= typemax(sample_type)
+            # XXX: we're being a bit clever here in order to not allocate a
+            # whole new sample array, plugging in the new sample_type, re-using
+            # the old encodoed samples data, and skipping validation.  this is
+            # okay in _this specific context_ since we know we're actually
+            # converting everything to Int16 in the actual export.
```
palday, line 143:
```suggestion
            # converting everything to sample_type in the actual export.
```

src/export_edf.jl
```diff
@@ -118,7 +190,9 @@ function onda_samples_to_edf_signals(onda_samples::AbstractVector{<:Samples}, se
                                                  extrema.physical_min, extrema.physical_max,
                                                  extrema.digital_min, extrema.digital_max,
                                                  "", sample_count)
-            sample_data = vec(samples[channel_name, :].data)
+            # manually convert here in case we have input samples whose encoded
+            # values are convertible losslessly to Int16:
+            sample_data = Int16.(vec(samples[channel_name, :].data))
```
palday, line 195: this seems unlikely but I guess it doesn't hurt 🤷

palday (approved):

palday:
I kinda think that this should be a method of `Onda.encode_sample`. To that end, you might want to add a comment to the docstring that this is an internal function and could disappear in the future (e.g. if moved into Onda itself).

kleinschmidt:
> I kinda think that this should be a method of Onda.encode_sample. To that end, you might want to add a comment to the docstring that this is an internal function and could disappear in the future (e.g. if moved into Onda itself).

Hm, I'm not sure about this...it's not part of the Onda documented API so I'm a bit reluctant, and it's also not _really_ doing the same thing (IIRC, `Onda.encode_sample` is actually _doing_ the encoding given specified parameters; we're _choosing_ parameters here _and_ encoding a whole `Samples`, not just a single sample...)

src/export_edf.jl
kleinschmidt: yup, hence the `XXX` comment above, wanted to make sure this was known to be cursed

src/export_edf.jl
kleinschmidt: ah, this is actually necessary when you have input samples with `Int8`/`UInt8` encoding (which is the "mildly cursed" bit you called out above)

src/export_edf.jl
```diff
@@ -96,18 +96,90 @@ function onda_samples_to_edf_header(samples::AbstractVector{<:Samples};
...
+`Samples` will be encoded, with a `info.sample_type` that is either equal to
+`sample_type` or losslessly `convert`ible.
+
+If the `info.sample_type` of the input samples cannot be losslessly converted to
+`sample_type`, new quantization settings are chosen based on the actual signal
+extrema, choosing a resolution/offset that maps them to `typemin(sample_type),
+typemax(sample_type)`.
+
+Returns an encoded `Samples`, possibly with updated `info`.  If the current
+encoded values can be represented with `sample_type`, nothing is changed.  If
```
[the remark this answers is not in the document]
kleinschmidt: how about

```suggestion
encoded values can be represented with `sample_type`, the `.info` is not changed.  If
```

PR #77, opened 2023-06-29
match units, labels, and encoding independently while planning

### LINKED ISSUE #69: label matching should be decoupled from unit matching
If _either_ the units _or_ the labels for a signal fail to match, `missing`s are set for ALL the Onda fields, even if the other one would have matched.  This can make it really hard to figure out why matching has failed.

kleinschmidt:
A common issue that users experience is that when _either_ labels or units fail to match, _everything_ is set to missing (including the encoding parameters).  This makes it difficult to tell what the source of the failure to match is, and also makes it functionally impossible to do a manual "clean up" edit of the plan after the fact.

This PR hoists the unit and encoding matching out of the loop that matches labels.

Fixes #69

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OndaEDF"
 uuid = "e3ed2cd1-99bf-415e-bb8f-38f4b42a544e"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.11.6"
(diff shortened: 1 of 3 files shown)


josephsdavid (approved):
🙏 This is super helpful. Needs a version bump and then you are good to go!

PR #78, opened 2023-07-07
more narrative docs on converting EDF to Onda

kleinschmidt:
As part of some internal Beacon work we've been working on a playbook for converting EDFs to Onda.  We've decided that public OndaEDF docs are a good place for most of that, so I'm expanding the docs a bit more in this PR to cover in more detail our standard workflow for this conversion.

Will pull material out of the readme since it's now duplicated but want to get feedback on the content first!

---
  - https://app.asana.com/0/0/1205001403646385


docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+
+The second option you have is to provide custom `labels=` and `units=` keyword arguments to [`plan_edf_to_onda_samples`](@ref).
+For unambiguous, [spec-compliant](https://www.edfplus.info/specs/edftexts.html#label_physidim) `label`s and `physical_dimension`s, it's generally possible to create custom `label=` or `unit=` specifications to match them.
+
+!!! note
+    Custom labels should be specified as _lowercase_, without reference, and without the sensor type prefix.
+    So to match a label like `"EEG R1-Ref"`, use a label like `"eeg" => ["r1"]`, and not `"EEG" => ["R1"]` or `"eeg" => ["r1-ref"]`.
+    See the documentation for [`plan_edf_to_onda_samples`](@ref) for more details, and the internal [`OndaEDF.match_edf_label`](@ref) for low-level details of how labels are matched.
+
+#### Preprocessing signal headers
```
kimlaberinto, line 59: Great section here with examples

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+
+plans = map(plan_edf_to_onda_samples ∘ corrected_header, edf.signals)
+new_plan = plan_edf_to_onda_samples_groups(plans)
+```
+
+Note that an additional step of `plan_edf_to_onda_samples_groups` is required after planning the individual signals.
+This is due to the fact that EDF is a "single channel" format, where each signal is only a single channel, while Onda is a "multichannel" format where a signal can have mmultiple channels as long as the sampling rate, quantization, and other metadata are consistent.
+Normally, calling `plan_edf_to_onda_samples` with an `EDF.File` will do this grouping for you, but when planning individually pre-processed signal headers, we have to do it ourselves at the end.
+
+#### Modification of the generated plan
```
kimlaberinto, line 86: Also great section here with examples

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+new_plan = plan_edf_to_onda_samples_groups(new_plan)
+```
+
+### Execute the plan
+
+Once the plan has been reviewed and deemed satisfactory, execute the plan to generate `Onda.Samples` and an "executed plan" record.
+This is accomplished with the [`edf_to_onda_samples`](@ref) function, which takes an `EDF.File` and a plan as input, and returns a vector of `Onda.Samples` and the plan as executed.
+The executed plan may differ from the input plan.
+Most notably, if any errors were encountered during execution, they will be caught and the error and stacktrace will be stored as strings in the `error` field.
+It is important to review the executed plan a final time to ensure everything was converted as expected and no unexpected errors were encountered.
```
kimlaberinto, line 149: This is a great addition here

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+## Batch conversion of many EDFs
+
+The workflow for bulk conversion of multiple EDFs is similar to the workflow for converting a single EDF.
+The major difference is that the "planning" steps can be conducted in bulk, while the "execution" steps (generally) need to be conducted one at a time, either serially or distributed across multiple workers.
+As discussed above, the planning stage requires only a few KB from the EDF file/signal headers, facilitating rapid plan-review-revise iteration of even fairly large collections of EDFs (10,000+).
+
+### Planning multiple EDFs
+
+The main factor to consider when planning conversion of a large batch of EDF files is that planning requires only the (small number) of header bytes, even for very large EDF files.
+Thus, the first step is to read the file headers into memory without reading the signal data itself (which for more than a few EDF files will not in general fit into memory ).
```
kimlaberinto, line 166: Nitpicky, I'm ok either way

```suggestion
Thus, the first step is to read the file headers into memory without reading the signal data itself (which for more than a few EDF files will not usually fit into memory due to the large amount of signal data found in EDF files).
```

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+end
+```
+
+#### Reading headers from S3
+
+!!! note
+    This section may become obsolete in a future version of EDF.jl which uses the [conditional dependency](https://pkgdocs.julialang.org/v1/creating-packages/#Weak-dependencies) functionality available from Julia 1.9+ to provide tighter integration with AWSS3.jl.
+
+Unfortunately, `open(path::S3Path)` will fetch the entire contents of the object stored at `path`, so we need to be a bit clever to read _only_ header bytes from an S3 file, especially given that the number of bytes we need to read depends on the number of signals.
+The following is an example of one technique for reading EDF file and signal headers from S3:
```
kimlaberinto, line 184: Wonderful addition with these examples

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+end
+
+function EDF.File(path::S3Path)
+    _, n_signals = EDF.read_file_header(path)
+    bytes = s3_get(path.bucket, path.key; byte_range=1:(256 * (n_signals + 1)))
+    return EDF.File(IOBuffer(bytes))
+end
+
+# use asyncmap because this is mostly bound by request roundtrip latency
+files = asyncmap(EDF.File, edf_paths)
```
kimlaberinto, line 200: Do we want to be a bit more general here and not use piracy(?) to define these functions, so that someone can easier copy and paste and use this?

e.g. (off the top of my head)
```suggestion
function _read_file_header(path::S3Path)
    bytes = s3_get(path.bucket, path.key; byte_range=1:256)
    buffer = IOBuffer(bytes)
    return EDF.read_file_header(buffer)
end

function _load_edf_file_header(path::S3Path)
    _, n_signals = _read_file_header(path)
    bytes = s3_get(path.bucket, path.key; byte_range=1:(256 * (n_signals + 1)))
    return EDF.File(IOBuffer(bytes))
end

# use asyncmap because this is mostly bound by request roundtrip latency
files = asyncmap(_load_edf_file_header, edf_paths)
```

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+
+unmatched_units = filter(:sample_unit => ismissing, plans)
+@info "unmatched labels:" tally(unmatched_units, :physical_dimension)
+
+matched = subset(plans, :channel => ByRow(!ismissing), :sample_unit => ByRow(!ismissing))
+@info "matched sensor types/channels:" tally(matched, [:sensor_type, :channel, :sample_unit])
+```
+
+Reviewing these summaries is a good first step when revising the plans.
+The revision process is basically the same as with a single EDF: udpate the `labels=` and `units=` as needed to capture any un-matched EDF signals, and failing that, preprocess the headers/postprocess the plan.
```
kimlaberinto, line 244:
```suggestion
The revision process is basically the same as with a single EDF: update the `labels=` and `units=` as needed to capture any un-matched EDF signals, and failing that, preprocess the headers/postprocess the plan.
```

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+        return insertcols!(plan, 
+                           :origin_uri => origin_uri,
+                           :recording => recording)
+    end
+end
+```
+
+### Review and revise the plans
+
+This "bulk plan" table can then be reviewed in bulk, looking for patterns in which `label`s are not matched, physical units associated with each `sensor_type`, etc.
```
kimlaberinto, line 226: Do we want to mention that determining custom labels is sometimes context dependent on what other signals are present in the EDF?

Might already be obvious, not sure how helpful it is to someone. I want thinking this specifically the context of bulk review of multiple EDFs. Sometimes the same EDF label can be interpreted as different things pending on the context and presence of other signal labels in the same EDF (I unfortunately don't have a specific example of this off the top of my head)

kimlaberinto (approved):
Looks really good! The way you formed the sections seems very logical to me and makes it easy to navigate. I added some misc comments and suggestions, but nothing blocking

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Philosophy
+
+The motivation for separating the planning and execution is twofold.
+
+First, making the plan a separate intermediate output means that not only can it be reviewed during conversion but can be persisted as a record of how any EDF-derived Onda signals were converted.
+This kind of provenance information is very useful when investigating issues with a dataset that may crop up long after the initial conversion.
+
+Second, planning only requires that the _headers_ of the `EDF.Signal`s be read into memory, thereby separating the iterative part of the conversion process from the expensive, one-time step which requires _all_ the signal data be read into memory.
+This enables workflows that would be impossible otherwise, like planning bulk conversion of thousands of EDFs at once.
```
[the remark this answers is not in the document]
kleinschmidt: gonna go with "impractical"

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+When dealing with large, messy datasets, we have found that "long-tail" metadata issues are inevitable and are better dealt with in bulk, and the plan-then-execute workflow enables users to deal with these issues all at once, save out the plan, and then distribute the actual conversion work to as many workers as necessary to execute it in a reasonable timeframe.
```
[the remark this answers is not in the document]
kleinschmidt: yeah, it's a bit uh cryptic lol.  What i"m getting at is that if there's one mis-labeled signal, then there are probably more of them, and if it's related to human error they're probably mis-labeled in different ways.  I'll spell it out more explicilty.

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+As outlined in the documentation for [`plan_edf_to_onda_samples`](@ref), a "plan" is a table with one row per `EDF.Signal`, which contains all the fields from the signal's header as well as the fields of the `Onda.SamplesInfoV2` that will be generated when the plan is executed (with the caveat that the `channels` field is called `channel` to indicate that it corresponds to a single channel in the output).
+
+### Review the plan.
+
+Check for EDF signals whose `label` or `physical_dimension` could not be matched using the standard OndaEDF labels and units, as indicated by `missing` values in the `channel`/`sensor_type` (for un-matched `label`) or `sample_unit` (for un-matched `physical_dimension`).
+It's also a good idea at this point to review the other EDF signal header fields, and how they will be converted to Onda (especially the sample unit, resolution and offset, which correspond to the physical/digital minimum/maximum from the EDF signal header.
+However, it's harder to fix issues with these fields, but it's still better to detect and document any issues with the underlying EDF data at this stage to prevent nasty surprises down the road.
```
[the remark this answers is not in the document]
kleinschmidt: great, much clearer!

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
...
+The first option to consider is to simply ignore these signals; not all signals are necessarily required for downstream use, and converting each and every signal in an EDF may be more work than is justified!
+
+#### Custom labels and units
+
+The second option you have is to provide custom `labels=` and `units=` keyword arguments to [`plan_edf_to_onda_samples`](@ref).
+For unambiguous, [spec-compliant](https://www.edfplus.info/specs/edftexts.html#label_physidim) `label`s and `physical_dimension`s, it's generally possible to create custom `label=` or `unit=` specifications to match them.
+
+!!! note
+    Custom labels should be specified as _lowercase_, without reference, and without the sensor type prefix.
+    So to match a label like `"EEG R1-Ref"`, use a label like `"eeg" => ["r1"]`, and not `"EEG" => ["R1"]` or `"eeg" => ["r1-ref"]`.
```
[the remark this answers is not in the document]
kleinschmidt: that's a great idea actually...

docs/src/convert-to-onda.md
kleinschmidt: yeah, I originally had something like this (in the internal code I'm borrowing from) but wanted to make this as close to what I'd put in the EDF.jl package extension as possible...

docs/src/convert-to-onda.md
kleinschmidt: yeah that's a very good point, I'll add a bit about that (and more caveats about how labels are matched), maybe even in the earlier section ...

docs/src/convert-to-onda.md
```diff
@@ -0,0 +1,264 @@
+# An Opinionated Guide to Converting EDFs to Onda
+
+## Basic workflow
+
+At a high level, the basic workflow for EDF-to-Onda conversion is iterative:
+1. Formulate a "plan" which specifies how to convert the metadata associated with each `EDF.Signal` into Onda metadata (channel names, quantization/encoding parameters, etc.)
+2. Review the plan, making sure that all necessary `EDF.Signal`s will be extracted, and that the quantization, sample rate, physical units, etc. are reasonable.
+3. Revise the plan as needed, repeating steps 1-2 until you're happy.
+4. Execute the plan, loading all EDF signal data if necessary and converting into `Onda.Samples`
```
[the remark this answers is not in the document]
kleinschmidt: added, this is a good point to flag up front!

PR #79, opened 2023-07-13
fix docs build on tags without explicitly adding OndaEDFSchemas

kleinschmidt:
The documenter build is failing on tags but not on the other builds.  I think this is because the other builds, we explicitly `dev` the local version of OndaEDFSchemas, but on tags we don't.  This PR fixes the docs build by using `OndaEDF.OndaEDFSchemas` instead of `OndaEDFSchema` directly.


palday (approved):
yeah, let's try this and see if it works 😄

PR #91, opened 2024-01-27
TimeSpans 1 compat

palday:


ericphanson (approved):
`==` isn't really expected for these roundtrips, and we added fma to decoding (but didn't touch encoding) so it's not totally surprising they varied a bit

PR #104, opened 2025-01-28
Support EDF.jl 0.8

haberdashPI:
This upgrades OndaEDF.jl to EDF.jl 0.8 (c.f. https://github.com/beacon-biosignals/EDF.jl/pull/92).

~~CI won't pass yet, since I'll need to merge https://github.com/beacon-biosignals/EDF.jl/pull/92 and register the package first.~~


kleinschmidt (commented):
also needs a change in the export code, here: https://github.com/beacon-biosignals/OndaEDF.jl/blob/aa3eca78b7ff15f90f9526cf800ff2ff783ae5fc/src/export_edf.jl#L26

OndaEDFSchemas.jl/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "OndaEDFSchemas"
 uuid = "9c87d999-769b-4741-85b2-6f554d09e731"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.2.2"
+version = "0.3.0"
```
ararslan, line 4: I don't think this is breaking since it's just adding new schemas that use the widened field type rather than changing any existing schemas.

haberdashPI: Right now this only supports EDF 0.8 (not 0.7 and earlier). This means that the returned value of a number of functions is FilePlanV3 / PlanV3 instead of V2.

haberdashPI: Oh I see, I somehow missed that this is for OndaEDFSchemas.

haberdashPI:
Alright, I did another search to make sure I didn't miss any other places where the type needs to be widened from `Int16` to `Int32`.

test/export.jl
```diff
@@ -31,7 +31,9 @@
     @testset "Record metadata" begin
         function change_sample_rate(samples; sample_rate)
             info = SamplesInfoV2(Tables.rowmerge(samples.info; sample_rate=sample_rate))
-            new_data = similar(samples.data, 0, Onda.index_from_time(sample_rate, Onda.duration(samples)) - 1)
+            # sparse CSC but for wide data we have a huge column pointer
+            # so transposing to get sparse CSR
+            new_data = spzeros(eltype(samples.data), Onda.index_from_time(sample_rate, Onda.duration(samples)) - 1, channel_count(samples))'
```
kleinschmidt, line 36: I don't understand this change; AFAICT it's not really related to the EDF.jl changes, but maybe I'm missing something?

haberdashPI: @palday did you mean to push this change here?

palday: tests [were failing on 1.6](https://github.com/beacon-biosignals/OndaEDF.jl/actions/runs/13020881507/job/36321044154#step:7:250) because whatever Onda version gets resolved there ignored `validate=false` when using the old 0d array. Sparse matrices avoid that and the allocation of a massive dense array.

kleinschmidt: we can drop 1.6 support IMO, it's not longer LTS

palday: I'm also fine with that, whether or not you leave the sparse change in.

haberdashPI: Turns out this is still a problem in 1.10, so I'm going to keep @palday's change.

palday: I'm so sorry. I really didn't mean to bring this upon you!

haberdashPI: 😆

palday (approved):
Approving to get things unblocked elsewhere: the changes make sense and I don't think we missed anything. If we did, that's what patch releases are for. 😂

Issue #110, opened 2025-04-24
how to use `sensor_label`?


ericphanson:
this `Onda.store` invocation does not set it:

https://github.com/beacon-biosignals/OndaEDF.jl/blob/a32ffe5fe80aa5c90dea6e6e3f7679f93205881c/src/import_edf.jl#L754-L759

I'm not using that one, I'm using `edf_to_onda_samples` which returns a list of samples and plans, but one of them is sorted so they aren't in sync, so I can't find the appropriate `sensor_label` entry of the plan to set in my own `Onda.store` invocation:

https://github.com/beacon-biosignals/OndaEDF.jl/blob/a32ffe5fe80aa5c90dea6e6e3f7679f93205881c/src/import_edf.jl#L544-L572

kleinschmidt:
The sort is to ensure that the order is consistent, even if things are skipped:

https://github.com/beacon-biosignals/OndaEDF.jl/blob/a32ffe5fe80aa5c90dea6e6e3f7679f93205881c/src/import_edf.jl#L516-L517

you _should_ be able to align based on that.

But OndaEDF isn't actually setting the sensor label; that's ahem left as an exercise to the reader 😁

But here's more or less how I'd go about it

```julia
type_counts = Dict{String,Int}()
for samples in list_of_samples
    (; sensor_type) = samples.info
    count = get!(type_counts, sensor_type, 0) + 1
    sensor_label = count > 1 ? sensor_type * '_' * count : sensor_type
    Onda.store(...)
    type_counts[sensor_type] = count
end
```

I agree that the finding the alignment between the returned samples and the plan rows is a bit annoying; the way to do it is to use the group idx skip any groups where any of the samplesinfo-relevant quantities is `missing`.

ericphanson:
hm, that could work. We modify the plan to have a consistent/correct/unique sensor label, so the frustrating bit is that we have the info we need in the `plan`, but we get back `samples` and `executed_plans` which are out of sync with each other so we can't just `for (samples, plan) in zip(all_samples, all_plans)` ... and use `plan.sensor_label`. So we need to re-divine a sensor-label out of `samples` alone, even though the correct sensor label is already in the plan.

kleinschmidt:
Fixed w/ #112

PR #112, opened 2025-04-30
Use sensor_label as grouping in file plan and return ConvertedSamples struct

### LINKED ISSUE #106: OndaEDF docstrings refer to Onda "kind" instead of `sensor_type`
https://beacon-biosignals.github.io/OndaEDF.jl/stable/api/#OndaEDF.plan_edf_to_onda_samples

kleinschmidt:
This addresses #110.  The general approach is that the collection of Onda
Samples returned by `edf_to_onda_samples` is a ~~_dictionary_ that's keyed by a
string that's suitable for using as a `sensor_label`~~ an iterable of (successful and failed) 
`ConvertedSamples` which reflect both successful and failed conversions of groups of
EDF Signals to Onda.Samples and capture the plan rows and `sensor_label` for the
group as well as the actual samples (which will be `missing` if conversion failed for any reason).

`ConvertedSamples` can be `Onda.store`d which _propagates `missing` samples from 
failed conversions_, and, if given a recording + start `Period`, will
auto-propagate the `sensor_label`.

There are two convenience functions for getting the old-style split output from
the `Vector{ConvertedSamples}`: `get_samples` gets all teh successful samples and 
`get_plan` gets the entire executed plan as one table.

Under the hood this is accomplished by
replacing the `onda_signal_index` field from the `FilePlanV3` with a
`sensor_label` field in `FilePlanV4`, and removing `sensor_label` from `PlanV4`.
This is a breaking change but addresses a number of issues.

- `sensor_label` does not belong at the Plan level (it cannot be determined by
  information in a single EDF Signal alone) but at the FilePlan level (it must
  be unique within a recording, and so depends on other rows).
- Currently, modifying the `sensor_label` in the generated plans has no effect on the
  execution of the plan.
- `onda_signal_index` is a lie: some groups may be skipped so while the
  _ordering_ of the Samples output is consistent with the `onda_signal_index`
  values, the exact indices are not.

Fixes #106, #110

TODO: 
- [x] update documentation

---
  - https://app.asana.com/0/0/1210103396066561


docs/src/convert-to-onda.md
```diff
@@ -199,8 +200,57 @@ If any errors _were_ encountered, you may need to iterate further.
...
+    Legolas.record_merge(plan; recording, sensor_label="edf_" * plan.sensor_label)
+end
+
+edf_samples, plan_executed = edf_to_onda_samples(edf, plans)
+
+start_relative_to_recording = Second(0)
+edf_signals = map(keys(edf_samples)) do sensor_label
+    samples = edf_samples[sensor_label]
+    return Onda.store(joinpath("samples", sensor_label * ".lpcm"), "lpcm", samples,
+                      recording, start_relative_to_recording, sensor_label)
```
ericphanson, line 239: is there some way to get at the plan inside here? it feels unsatisfying if we only fix the sensor label thing without fixing the mismatch that we have 2 paired outputs but you can't associate them with each other

ericphanson: I think the way to do it now is filter `plan_executed` to matching sensor labels. which is alright, at least we can do it now

src/import_edf.jl
```diff
@@ -558,20 +569,21 @@ function edf_to_onda_samples(edf::EDF.File, plan_table; validate=true, dither_st
...
-    sort!(exec_rows; by=(row -> row.idx))
     exec = Tables.columntable(exec_rows)
-
     exec_plan = reduce(vcat, exec.plan_rows)
 
-    return collect(skipmissing(exec.samples)), exec_plan
+    samples = OrderedDict(key.sensor_label => samples
+                          for (; key, samples) in exec_rows if !ismissing(samples))
+
+    return samples, exec_plan
```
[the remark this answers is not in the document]
omus: We discussed this during a meeting and my thoughts on this are as follows:

- The intent (as I understand it) is that we return `samples` to make it easy to write out the data to disk while the plan is returned as modification may have occurred.
- The design decision around having the API of OndaEDF.jl perform the data transformation entirely in memory is correct in my opinion.
- The steps here which modify the plan should be encapsulated in a single function and the result of that function should be the modified plan (or a mutating function if practical).
- Returning a custom Julia type seems to be the right approach so we can return all of the information we want to relay to the user and then the user can use that type to easily extract the valid samples via a function provided by this package.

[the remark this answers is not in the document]
kleinschmidt: Yeah after discussing this with the team I'm on board with using a struct for this.  I agree with @omus point that the most Julian way to handle the coupling between different bits of information is a struct, because you can dispatch on that.  For instance, we could have a method for `Onda.store(path, format, ::ConvertedSamples, ...)` that populates the sensor label (and possibly handles the possibility of missing samples) etc.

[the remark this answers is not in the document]
kleinschmidt: Also a method for `write_plan` which we already define that concatenates the plans.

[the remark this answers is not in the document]
ericphanson: fwiw I think `write_plan` is not used much by beaconeers because our data map utils package wants the raw table. So we should make getting the concatenated table easy; whether or not we add the helper around `Legolas.write` doesn't seem important to me.

src/import_edf.jl
```diff
@@ -505,14 +508,84 @@ grouper(vars::AbstractVector{Symbol}) = grouper((vars..., ))
...
+    end
+end
+
+# TODO: return a signal extension?
+function Onda.store(file_path, file_format, samples::ConvertedSamples, recording, start,
+                    sensor_label=samples.sensor_label)
+    return if ismissing(samples.samples)
+        missing
+    else
+        Onda.store(file_path, file_format, samples.samples, recording, start, file_format)
```
ericphanson, line 604: missing `sensor_label`?

Project.toml
```diff
@@ -1,29 +1,33 @@
 name = "OndaEDF"
 uuid = "e3ed2cd1-99bf-415e-bb8f-38f4b42a544e"
 authors = ["Beacon Biosignals, Inc."]
-version = "0.13.2"
+version = "0.14.0"
```
omus, line 4: Don't want to commit to a 1.0.0?

Project.toml
```diff
@@ -1,29 +1,33 @@
...
 
 [deps]
+ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
```
omus, line 9: Should add a compat bound for this. Looks like we need to add Aqua.jl

codecov.yml
```diff
@@ -1,10 +1,14 @@
+coverage:
```
omus, line 1:
```suggestion
---
coverage:
```

docs/src/convert-to-onda.md
```diff
@@ -37,8 +37,9 @@ After the detailed workflow for converting a single EDF file to Onda format, we'
 This is straightforward, using [`plan_edf_to_onda_samples`](@ref).
 As outlined in the documentation for [`plan_edf_to_onda_samples`](@ref), a "plan" is a table with one row per `EDF.Signal`, which contains all the fields from the signal's header as well as the fields of the `Onda.SamplesInfoV2` that will be generated when the plan is executed (with the caveat that the `:channels` field is called `:channel` to indicate that it corresponds to a single channel in the output).
 It also contains a few additional fields for defining the mapping between EDF and Onda signal indices, as well as a field to capture any errors thrown during planning (or, more likely, during execution of the plan):
-- `:edf_signal_index`, the 1-based numerical index of the source signal in `edf.signals`
-- `:onda_signal_index`, the ordinal index of the resulting samples (not necessarily the index into `samples`, since some groups might be skipped)
+- `:edf_signal_index`, the 1-based numerical index of the source signal in `edf.signals`.
+- `:sensor_label`, a file-unique unique label for this group of channels (may be `missing`).
```
omus, line 41: We might want to be clearer on what "file-unique" means
```suggestion
- `:sensor_label`, a per EDF file unique label for this group of channels (may be `missing`).
```

docs/src/convert-to-onda.md
```diff
@@ -190,17 +191,96 @@ new_plan = plan_edf_to_onda_samples_groups(new_plan)
 ### Execute the plan
 
 Once the plan has been reviewed and deemed satisfactory, execute the plan to generate `Onda.Samples` and an "executed plan" record.
-This is accomplished with the [`edf_to_onda_samples`](@ref) function, which takes an `EDF.File` and a plan as input, and returns a vector of `Onda.Samples` and the plan as executed.
-The executed plan may differ from the input plan.
-Most notably, if any errors were encountered during execution, they will be caught and the error and stacktrace will be stored as strings in the `error` field.
-It is important to review the executed plan a final time to ensure everything was converted as expected and no unexpected errors were encountered.
+This is accomplished with the [`edf_to_onda_samples`](@ref) function, which takes an `EDF.File` and a plan as input, and returns a `Vector{ConvertedSamples}`.
```
omus, line 194: Just a thought. Since the function is `edf_to_onda_samples` it may make more sense for the struct to be named `ConvertedOndaSamples`

docs/src/convert-to-onda.md
```diff
@@ -190,17 +191,96 @@ new_plan = plan_edf_to_onda_samples_groups(new_plan)
...
+For each `sensor_label` in the plan, the corresponding [`ConvertedSamples`](@ref) captures the plan rows, the resulting `Onda.Samples`, and the `sensor_label`.
+
+It is important to review the converted samples a final time to ensure everything was converted as expected and no unexpected errors were encountered.
+First, check for any converted samples with `missing` values in the `.samples` field.
+These indicated that conversion of these plan rows failed for some reason, either because the EDF signal metadata could not be matched with appropriate Onda metadata or because of a runtime error during conversion.
```
omus, line 199: Seems like we should mention the `error` field here:
```suggestion
These indicated that conversion of these plan rows failed for some reason, either because the EDF signal metadata could not be matched with appropriate Onda metadata or because of a runtime error during conversion. If any errors were encountered during execution, they will be caught and the error and stacktrace will be stored as strings in the `error` field.
```

docs/src/convert-to-onda.md
```diff
@@ -190,17 +191,96 @@ new_plan = plan_edf_to_onda_samples_groups(new_plan)
...
-This is accomplished with the [`edf_to_onda_samples`](@ref) function, which takes an `EDF.File` and a plan as input, and returns a vector of `Onda.Samples` and the plan as executed.
-The executed plan may differ from the input plan.
-Most notably, if any errors were encountered during execution, they will be caught and the error and stacktrace will be stored as strings in the `error` field.
-It is important to review the executed plan a final time to ensure everything was converted as expected and no unexpected errors were encountered.
+This is accomplished with the [`edf_to_onda_samples`](@ref) function, which takes an `EDF.File` and a plan as input, and returns a `Vector{ConvertedSamples}`.
+For each `sensor_label` in the plan, the corresponding [`ConvertedSamples`](@ref) captures the plan rows, the resulting `Onda.Samples`, and the `sensor_label`.
+
+It is important to review the converted samples a final time to ensure everything was converted as expected and no unexpected errors were encountered.
+First, check for any converted samples with `missing` values in the `.samples` field.
```
omus, line 198: Consistency
```suggestion
First, check for any converted samples with `missing` values in the `samples` field.
```

omus (commented):
Partial review

docs/src/convert-to-onda.md
```diff
@@ -190,17 +191,96 @@ new_plan = plan_edf_to_onda_samples_groups(new_plan)
...
+OndaEDF.jl provides methods for `Onda.store` that accept [`ConvertedSamples`](@ref) and propagate `missing` samples.
+
+Note that `Onda.store` requires at a minimum the file path (or "path-like" writeable), the format to use to serialize the samples (like `"lpcm"` or `"lpcm.zst"`), and the `Samples` themselves.
+However, this only _serializes_ the samples to the file path.
+In most cases, you will want a _signal record_ that can be stored elsewhere, which combines the path and file format to the serialized samples along with the `SamplesInfo` and other metadata required to be able to _deserialize_ and use these samples.
+In this case, you should use the second `Onda.store` method which additionally requires:
+
+- `recording::UUID` (the recording that this signal is associated with)
+- `start::Period` (the start time of this signal _relative to the start of the recording_)
+- `sensor_label::AbstractString` (a _unique label_, defaults to the the `sensor_label` from the `ConvertedSamples`)
```
omus, line 221:
```suggestion
- `recording::UUID`: the recording that this signal is associated with
- `start::Period`: the start time of this signal _relative to the start of the recording_
- `sensor_label::AbstractString`: a _unique label_, defaults to the the `sensor_label` from the `ConvertedSamples`
```

omus (approved):

docs/src/convert-to-onda.md
kleinschmidt: I think the name (which was @ericphanson suggestion originally) comes from the fact that it wraps a `Samples` object.  could make the same argument for the function name too though but I'd rather keep it more concise as long as it's clear (and I think it is...)

.github/workflows/Documenter.yaml
```diff
@@ -0,0 +1,68 @@
+name: Documenter
```
kleinschmidt, line 1:
```suggestion
---
name: Documenter
```

docs/src/convert-to-onda.md
```diff
@@ -190,17 +191,96 @@ new_plan = plan_edf_to_onda_samples_groups(new_plan)
...
+Legolas.write(string(recording, ".ondaedf.file-plan.arrow"),
+              OndaEDF.FilePlanV4SchemaVersion(),
+              OndaEDF.get_plan(converted_samples))
+```
+
+#### Example: generating unique `sensor_label`s
+
+Consider a case with parallel recordings from a traditional EEG system (received as an EDF) and some custom EEG hardware (which emits Onda-formatted data natively).
+Each device generates multichannel EEG data and uses `sensor_type="eeg"`.
+We must adjust the generated sensor labels for the EDF signals before storing them so that they are unique, which we can do by prepending `"edf_"` to the default, `sensor_type`-derived labels before storing:
```
ericphanson, line 253: we don't do this here! also I think we should suggest modifying the plans rather than post-conversion

ericphanson (approved):
overall LGTM, but:

- would be good to get those `Onda.store` methods tested (or deleted if we want folks to use `get_samples`)
- while the "plan sensor_labels are respected" test is nice, it would be good to also test that duplicate labels yield unique sensor labels during planning
- looks like one bit of the new docs is out-dated, I left a comment there

approving since no re-review needed after those changes are made

docs/src/convert-to-onda.md
kleinschmidt: that's what these lines are doing: https://github.com/beacon-biosignals/OndaEDF.jl/blob/dd97c77d0912f20b77a5beb1e5971058f85342a7/docs/src/convert-to-onda.md?plain=1#L260-L262

kleinschmidt: I tried to separate the planning and modifying stages to make it clearer waht was going on , and added some comments.

PR #114, opened 2026-02-02
Add formatting script and call it as a pre-commit hook

MKVatinyan:
Basis taken from https://github.com/beacon-biosignals/OndaEDF.jl/pull/89, opted for a precommit hook instead of a CI with reviewdog

EDIT I added the CI but without the review dog stuff

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,8 @@ repos:
     hooks:
       - id: julia-format
         name: Julia formatter
-        entry: julia --project=format format/run.jl
-        language: system
+        entry: format/run.jl
+        language: julia # note: use pre-commit v4.3+
         files: \.jl$
+        additional_dependencies:
+          - 'JuliaFormatter@2'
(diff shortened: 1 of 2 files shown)


src/export_edf.jl
```diff
@@ -64,13 +69,13 @@ function Base.showerror(io::IO, exception::RecordSizeException)
...
     print(io, " cannot be resolved to a data record size smaller than ")
-    print(io, DATA_RECORD_SIZE_LIMIT * 2, " bytes")
+    return print(io, DATA_RECORD_SIZE_LIMIT * 2, " bytes")
 end
 
 function Base.showerror(io::IO, exception::EDFPrecisionError)
     print(io, "EDFPrecisionError: String representation of value ")
     print(io, exception.value)
-    print(io, " is longer than 8 ASCII characters")
+    return print(io, " is longer than 8 ASCII characters")
```
kleinschmidt:
```suggestion
    print(io, " is longer than 8 ASCII characters")
    return nothing
```

